PR #10, opened 2013-11-02
add has_auto_menu_dep_on for class Symbol, Choice, Menu and Comment

fpemud:
After add these functions, I successfully get a menu structure that is nearly the same as "make menuconfig".
For detail please check https://github.com/fpemud/fpemud-kernelmanager/blob/master/lib/kmenuconfiglib.py


ulfalizer:
Your `get_prompt()` and `get_selects()` functions actually make a lot of sense. Sorry for not looking hard enough earlier :(. It's nice that they take the conditions into account.

I might pull those (though the API is becoming a bit bloated again... wish I could redo it from scratch), and I'll add a thanks in that case too.

ulfalizer:
What's really needed is some general way to work with expressions though. If clients could fetch the various expressions in some format and deconstruct them (expressions can be arbitrarily complex), then they should be able to do "everything" after that, though not trivially in all cases. Helper functions could be added too.

I probably shouldn't add stuff like that until there's a strong demand though, as it might involve quite a lot of work and testing.

PR #13, opened 2015-02-16
Warn about and ignore the "allnoconfig_y" Kconfig option

jsitnicki:
A new option has appeared in Linux 3.15-rc1. Ignore it for now.

### DIFF
--- a/kconfiglib.py
+++ b/kconfiglib.py
@@ -1324,6 +1324,11 @@ error, and you should e-mail kconfiglib@gmail.com.
                                filename,
                                linenr)
 
+                elif tokens.check(T_ALLNOCONFIG_Y):
+                    self._warn("the 'allnoconfig_y' option is not supported.",
+                               filename,
+                               linenr)
+
                 else:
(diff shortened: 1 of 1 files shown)


ulfalizer:
Looks good. Thanks!

Ideally it should be made queryable, with the allnoconfig examples and tests updated to use it.

ulfalizer:
BTW -

It looks like the code has bitrotted a bit due to changes to Kconfig in the last few years. Plenty of FAILs in the test suite now.

The fact that everything still parses and all but one of the allyesconfig tests passes makes me think it's probably something minor (the test suite compares output character by character, so even really small and insignificant stuff can produce a FAIL). Also, it would likely only matter if you're using Kconfiglib to generate .config files.

I'll see if I can find some time to make all tests pass again. The current state of the test suite isn't very confidence-inducing. :)

jsitnicki:
Thanks for the quick merge and the library, in the first place. You're quite right, I've been playing only with the parser and not generating any .config's thus far. Thought it would be nice for the parser not to bail out until the generator gets updated.

PR #14, opened 2015-03-12
Fixes to make testsuite pass

philipc:
Tested on linux 3.19.

### DIFF
--- a/examples/allnoconfig.py
+++ b/examples/allnoconfig.py
@@ -13,9 +13,15 @@ while not done:
     done = True
 
     for sym in conf:
+        if sym.is_allnoconfig_y():
+            if sym.get_value() != 'y':
+                sym.set_user_value('y')
+                # We just changed the value of some symbol. As this may affect
+                # other symbols, keep going.
+                done = False
         # Choices take care of themselves for allnoconfig, so we only need to
         # worry about non-choice symbols
-        if not sym.is_choice_symbol():
+        elif not sym.is_choice_symbol():
             # If we can assign a value to the symbol (where "n", "m" and "y"
             # are ordered from lowest to highest), then assign the lowest
             # value. lower_bound() returns None for symbols whose values cannot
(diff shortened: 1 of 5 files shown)


ulfalizer:
Wonderful! Thanks a lot! :)

Sorry for lingering so long! Must've taken a while to figure out how things fit together and how to track down errors through diffing. :/

I get a hang in in *examples/allnoconfig.py* when running the testsuite with the latest kernel from https://github.com/torvalds/linux (**cca28a5fda5d8fa**, with tag **v4.0-rc3**). Don't worry about it though -- I'll fix it tomorrow. Here's an analysis which also illustrates some Kconfigness in case you're interested:

I think the problem is that `allnoconfig_y` is set on some symbol whose visibility prevents it from being `y`. For example, if `IS_ALLNOCONFIG_Y` depends on `DEP` and `DEP` is `n`, then the visibility of `IS_ALLNOCONFIG_Y` is `n`. The visibility of a symbol acts as an upper bound on the (actual) value, so setting the user value on `IS_ALLNOCONFIG_Y` to `y` won't actually change the (actual) value in that case. Such a symbol will cause the following to loop forever, since `sym.get_value()` will never be `y`:

    if sym.is_allnoconfig_y():
        if sym.get_value() != 'y':
            sym.set_user_value('y')
            # We just changed the value of some symbol. As this may affect
            # other symbols, keep going.
            done = False

The same problem will occur with `allnoconfig_y` on symbols whose visibility is `m` (which only allows them to be included as modules).

I'll either merge your commits as-is and add a separate fix for that problem, or at least provide an attribution in the README. Thanks!

philipc:
Thanks for fixing my changes :) Your explanation and changes make sense, except I don't get the hang when I try it, and the only symbol with the allnoconfig_y option doesn't have any dependencies. Not an issue, just trying to make sure I understand :)

ulfalizer:
Not even for v4.0-rc3?

Hmm... yeah, `EMBEDDED` does seems to be the only symbol with `option allnoconfig_y`, and it doesn't have any dependencies for `x86_64` at least. Not sure what's going off the top of my head to be honest. It might be that some other arch adds additional dependencies, though that'd be weird too. Or maybe I messed up somehow if you don't get the hang.

Might look into it some more later. Weird stuff. :)

PR #23, opened 2016-10-06
Extensible kconfig

kaizouman:
Hi,

These are two patches that introduce minor additions to the "standard" Kconfig language to allow a source tree to be more easily extended, following a syntax similar to GNU Make.

I also have patches for the "legacy" Kconfig implementation in the Linux Kernel, but I think it would be easier to integrate them in your parallel project.

### DIFF
--- a/kconfiglib.py
+++ b/kconfiglib.py
@@ -69,6 +69,7 @@ might add it in a safe way as a client API instead."""
 import os
 import re
 import sys
+import glob
 
 # File layout:
 #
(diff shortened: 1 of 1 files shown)


ulfalizer:
Hello,

I think I prefer to have Kconfiglib just emulate the behavior of the C implementation and not include any language extensions at this point. This is not a high-activity project though, so maintaining the patch separately should be trivial.

The following hack (along with the code in `Config._tokenize()` that uses it) could probably be removed now btw. It was originally needed to emulate sloppiness in the C lexer that some Kconfig files unknowingly relied upon, but the lexing has been improved since then. I only keep it around for backwards compatibility. 

    _initial_token_re_match = re.compile(r"[^\w]*(\w+)\s*").match

kaizouman:
OK. I will maintain these in my clone.

PR #25, opened 2017-02-13
support for CONFIG_ environment variable and minor change to README

cdornsife:
I used your project recently and needed to change CONFIG_ in the env. I added support so it uses CONFIG_ from the environment if it is set.

```
# https://kernel.org/doc/Documentation/kbuild/kconfig.txt
CONFIG_
--------------------------------------------------
If you set CONFIG_ in the environment, Kconfig will prefix all symbols with its value when
saving the configuration, instead of using the default, "CONFIG_".
```

Hope you find this PR useful.


ulfalizer:
Merged - thanks!

I modified the commit message a bit. The usual convention for Git is to have the summary sound like a command (imperative). I hope that's okay.

PR #35, opened 2018-01-04
Implement autoconf.h generation

carlescufi:
Implement the generation of the C header file that mirrors the .config
files, commonly named autoconf.h.

Signed-off-by: Carles Cufi <carles.cufi@nordicsemi.no>


kconfiglib.py
```diff
@@ -826,6 +826,69 @@ def _load_config(self, filename, replace):
...
+            node = self.top_node.list
+            if not node:
+                # Empty configuration
+                return
+
+            while 1:
+                item = node.item
+                if isinstance(item, Symbol):
+                    if not item._written:
+                        if item._write_to_conf:
```
ulfalizer: There's a subtlety here that causes this to break unless all the symbol values have already been calculated: `_write_to_conf` is calculated when the symbol value is calculated.

If you do `val = item.str_value` before checking `_write_to_conf` and use that, it will work fine. `.str_value` is actually a hidden function call (that uses Python `@property` magic).

There's a related note in a comment in `config_string`, but it might be hard to spot unless you already know that that's a hidden function call too. :)

kconfiglib.py
```diff
@@ -826,6 +826,69 @@ def _load_config(self, filename, replace):
...
+                            if item.orig_type in (BOOL, TRISTATE) and item.str_value != "n":
+                                suffix = "_MODULE" if item.str_value == "m" else ""
+                                write("#define {}{}{} 1\n"
+                                      .format(self.config_prefix, item.name,
+                                              suffix))
+                            elif item.orig_type == STRING:
+                                write('#define {}{} "{}"\n'
+                                      .format(self.config_prefix, item.name,
+                                              item.str_value))
```
ulfalizer: Could make this `kconfiglib.escape(item.str_value)`. Then string values containing `"` and `\` will be properly escaped.

kconfiglib.py
```diff
@@ -826,6 +826,69 @@ def _load_config(self, filename, replace):
...
+                            elif item.orig_type in (INT, HEX):
+                                write("#define {}{} {}\n"
+                                      .format(self.config_prefix, item.name,
+                                              item.str_value))
```
ulfalizer: Could do this, just to match the C implementation:

```python
if not val.startswith(("0x", "0X")):
    val = "0x" + val
```

Looks fine other than that!

I'm used to Gerrit. Think you might have to force push the branch or something to address review comments. :thinking:

carlescufi: Yeah, I do it all the time with GitHub :)
Addressed but I couldn't keep it within the 80-char width limit, Python was throwing me errors.

kconfiglib.py
carlescufi: Fixed thanks!

kconfiglib.py
ulfalizer: No worries.

If you want to split something that's not enclosed in brackets over multiple lines in Python, you need to add a \ at the end of the line.

carlescufi:
Thanks!

ulfalizer:
Going to bed soon (my sleep schedule's a bit wonky atm :)), but I'll push out a new release to PyPI later on today as well.

carlescufi:
Thank you, although there is no rush yet with PyPI since we will need to keep 2 patches (globbing and prefer later defaults) out of tree until I have fixed our Kconfig files.

Issue #38, opened 2018-02-25
Proposal: support relative path in 'source' statement


RomaVis:
Hi,

I'm working on a project where build configuration system is implemented entirely in Python with Kconfiglib at its center. That way, we don't have to use any of Linux kernel Kconfig/Kbuild code, which has certain benefits. In our situation it's tempting to extend Kconfiglib with new features beyond what's offered by Linux kernel, and one such extension that may be useful for someone else is the support for relative paths in `source` directive.

Imagine the following directory structure:
```
Project
|
+--Kconfig
+--src
   +--Kconfig
   |
   +--SubSystem1
   |  +--Kconfig
   |  |
   |  +--ModuleA
   |  |  +--Kconfig
   |  |
   |  +--ModuleB
   |     +--Kconfig   
   |
   +--SubSystem2
      +--Kconfig
      |
      +--ModuleC
      |  +--Kconfig
      |
      +--ModuleD
         +--Kconfig
```
Here, Kconfig file in `SubSystem1` sources Kconfig file from `ModuleA`, and to do this one writes:
```
source "src/SubSystem1/ModuleA/Kconfig"
```

Complete path from project root needs to be specified, which is error-prone and creates difficulties when for some reason one wants to rename/move `SubSystem1` directory, as paths should be modified in all Kconfig files under affected directory. Sometimes Kconfig files are nested more deeply than in my example, so it becomes even worse.

The proposal is to add support for relative paths in `source` directive, so that path relative to directory of currently processed Kconfig file can be specified instead of path relative to project root:
```
source "ModuleA/Kconfig"
```

Obviously, this mode of `source` directive interpretation is not compatible with Linux kernel tools. I see two ways how this can be enabled without breaking existing code:
1. Add optional argument to Kconfig constructor:
   ```
   def __init__(self, filename="Kconfig", warn=True, source_uses_relative_paths=False):
   ```
   This doesn't change Kconfig language, but in many other aspects it's ugly:
   - If such Kconfig infrastructure is by mistake used with Linux kernel tools, cryptic error messages would be rised about non-existing files. Or, even worse, if relative path specified in Kconfig file nested deep inside project directory tree _exists_ in project root, then Linux tools may interpret configuration incorrectly without issuing warnings of any kind..
   - The support for new constructor argument needs to be implemented in all tools based on Kconfiglib that you want to use in your configuration system
   - It's implicit: you can't immediately tell by looking into Kconfig files which kind of paths is used
[...]

PR #39, opened 2018-02-26
Implement 'rsource' statement ('source' with relative path)

RomaVis:
Allows sourcing Kconfig files from subdirectories via relative path with new 'rsource' Kconfig statement.
Added some documentation to module docstring and simple test in testsuite.py.

For more details check Issue #38


kconfiglib.py
```diff
@@ -4452,11 +4513,12 @@ def _check_choice_sanity(choice):
     _T_RANGE,
     _T_SELECT,
     _T_SOURCE,
+    _T_RSOURCE,
```
ulfalizer: This list is sorted alphabetically.

kconfiglib.py
```diff
@@ -4492,6 +4554,7 @@ def _check_choice_sanity(choice):
     "range":          _T_RANGE,
     "select":         _T_SELECT,
     "source":         _T_SOURCE,
+    "rsource":        _T_RSOURCE,
```
ulfalizer: Alphabetically here too.

kconfiglib.py
```diff
@@ -4509,6 +4572,7 @@ def _check_choice_sanity(choice):
     _T_MENU,
     _T_PROMPT,
     _T_SOURCE,
+    _T_RSOURCE,
```
ulfalizer: Alphabetically here too.

tests/include/Klocation_included
```diff
@@ -0,0 +1,6 @@
+
+comment "some comment"
+
+
+config MULTI_DEF
+    bool
```
ulfalizer: Thanks for adding some tests!

Comments:

 - There is already a `tests/sub/` directory. We could reuse that one instead of adding `tests/include/`.

 - Maybe `Klocation_included` could be renamed to `Klocation_sourced`. Then the new file could be called `Klocation_rsourced`.

 - I think `Klocation_rsourced` could have just `config MULTI_DEF` in it. A symbol defined in multiple locations doesn't need to specify the type at every location, and the comment seems to be unused.

 - Could you rename `EXPANDED_FROM_ENV` to `EXPANDED_IN_SOURCE` as well? Then `INCLUDE_DIR_FROM_ENV` could be called `EXPANDED_IN_RSOURCE`.

Thanks!

kconfiglib.py
```diff
@@ -1102,31 +1102,31 @@ def _open(self, filename):
...
+    def _translate_relative_path(self, path):
+        # We expect strictly relative path, not absolute
+        if os.path.isabs(path):
+            raise KconfigSyntaxError(
+                "\n{}:{}: Relative path expected, while '{}' is absolute.\n"
+                "Backtrace:\n{}"
+                    .format(self._filename, self._linenr, path,
+                            "\n".join("{}:{}".format(name, linenr)
+                                      for _, name, linenr
+                                      in reversed(self._filestack))))
```
ulfalizer: I think it would be better to remove this check and just specify that `rsource` works the same as `source` for absolute paths ("If an absolute path is given to 'source' or 'rsource', it will be opened directly, ignoring `$srctree` and (for 'rsource') the location of the current Kconfig file", or something like that, in the new section that describes `rsource`).

`os.path.join("foo", "/bar")` ignores `"foo"` and returns `"/bar"` btw, so `join()` will "just work" there.

After that, the logic will be so short that `_translate_relative_path()` could be removed.

ulfalizer: This might be slightly better still:

    Like for 'source', an absolute path given to `rsource` will be opened directly,
    ignoring $srctree and the location of the current Kconfig file.

kconfiglib.py
RomaVis: Missed that, fixed. Same for next two.

tests/include/Klocation_included
RomaVis: Implemented per your description, with an exception that I've used 'TESTS_DIR_FROM_ENV' and 'SUB_DIR_FROM_ENV' names for envvar and config symbols.

kconfiglib.py
RomaVis: Agree. Absolute paths allowed, `_translate_relative_path()` removed.

ulfalizer:
Looks good to me now. Thanks!

As one last change, could you squash the four commits into a single commit and make a new pull request?

(Just in case you usually work with some other VCS, see https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git. When pushing afterwards, you will need to use `--force`, as you're changing the history of the branch.)

RomaVis:
Thanks for review!

As you've said, I'm not primarily using git, so I have probably rather silly question: instead of rewriting my repo history, won't ["Squash and merge" PR option](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits) handle squashing automatically during merge?

ulfalizer:
I think you're right. Hadn't noticed that option.

I'll give it a go.

ulfalizer:
Seems to have worked. I learned something new.

PR #41, opened 2018-03-12
Decode sources as UTF-8 instead of relying on the system locale

SebastianBoe:
Opening files without specifying an encoding will cause the system
locale to determine the encoding. It is undesired behaviour IMHO that
reading/decoding Kconfig sources is platform-dependent, instead
kconfiglib should behave in the same way on all platforms.

Example:

$ export LANG=C
$ echo "# Author: Sebastian Bøe" >> Kconfig
$ test_kconfig_lib.py

Alternatively, one could decode as ascii, especially if it is
explicitly documented that Kconfig sources (including comments) must
be ascii.

The most important thing is that kconfiglib behaves the same on all
platforms.

Signed-off-by: Sebastian Bøe <sebastian.boe@nordicsemi.no>

### DIFF
--- a/kconfiglib.py
+++ b/kconfiglib.py
@@ -358,6 +358,7 @@ import os
 import platform
 import re
 import sys
+import io
 
 # File layout:
 #
@@ -1314,12 +1315,12 @@ class Kconfig(object):
         # was set when the configuration was loaded
 
         try:
-            return open(filename, _UNIVERSAL_NEWLINES_MODE)
+            return io.open(filename, _UNIVERSAL_NEWLINES_MODE, encoding='utf-8')
         except IOError as e:
             if not os.path.isabs(filename) and self.srctree is not None:
                 filename = os.path.join(self.srctree, filename)
                 try:
-                    return open(filename, _UNIVERSAL_NEWLINES_MODE)
+                    return io.open(filename, _UNIVERSAL_NEWLINES_MODE, encoding='utf-8')
                 except IOError as e2:
                     # This is needed for Python 3, because e2 is deleted after
                     # the try block:


ulfalizer:
Seems reasonable. Would need to use `io.open()` to stay compatible with both Python 2 and 3 though, as Python 2's `open()` doesn't take an `encoding` parameter.

See the comment near `_UNIVERSAL_NEWLINES_MODE` as well. I guess it might not be a huge deal though.

There's `codecs.open()` too, but that one doesn't support universal newlines mode.

ulfalizer:
Maybe `_UNIVERSAL_NEWLINES_MODE` becomes pointless with `io.open()` too. I suspect it turns into a no-op, as it already defaults to universal newlines mode.

I'm off to bed now (slightly wonky sleep schedule :P), but I'll do some measuring tomorrow.

SebastianBoe:
You are correct, io.open defaults to universal newlines mode io (Py2 and Py3). So it is not necessary to explicitly enable it.

ulfalizer:
@mgomersbach
Forcing UTF-8 decoding would mean that files parse the same for everyone, regardless of their locale settings, which might be nice. The only downside I can think of is that people might be editing files using an encoding from some other locale and expect Kconfiglib to use the same one.

I live in relatively painless latin-1 land, and I'm not an i18n expert, so please tell me if you can think of cases where this behavior would get painful.

pfalcon:
+1 for this change. We recently had this issue that kconfiglib failed on parsing Kconfig files in Zephyr project (https://github.com/zephyrproject-rtos/zephyr). There're number of failures in public CI systems due to this, e.g.:

* https://ci.linaro.org/job/lite-aeolus-jerryscript/PLATFORM=frdm_k64f,ZEPHYR_GCC_VARIANT=zephyr,label=docker-xenial-amd64-13/257/console
* https://ci.linaro.org/job/lite-aeolus-micropython/PLATFORM=disco_l475_iot1,ZEPHYR_GCC_VARIANT=zephyr,label=docker-xenial-amd64-13/314/display/redirect

That issue was worked around on Zephyr side: https://github.com/zephyrproject-rtos/zephyr/pull/6599 , but as can be seen, the issue happened in author's name (copyright notice). It's unreasonable to expect everyone to change their name (or request them) to ASCII only. UTF-8 is both de jure and de facto universal encoding, so it only makes sense to use it by default.

Thanks!

ulfalizer:
[PEP 538](https://www.python.org/dev/peps/pep-0538/), implemented in the upcoming Python 3.7, is related. It specifically mentions Docker.

So, I think the problem could be fixed by specifying a UTF-8 locale when starting the container. At the same time, it would be nice if things just worked for people.

I guess a more "proper" fix might be to do something similar to what that PEP does, and use `encoding="utf_8"` if the `C` locale is detected. Not sure I trust myself with implementing that without more platforms to test it on though. I guess forcing UTF-8 everywhere might be okay for now.

ulfalizer:
Just noticed some more iffiness for Python 2.

Passing `encoding="utf-8"` will make `readline()` return a `unicode` string. That causes problems for checks like `if isinstance(foo, str)`, as `isinstance(u'foo', str)` is `False`. Python 2 also has `basestring`, which can be used to check for either string type, but it's not available in Python 3.

It could be worked around, but I'm not sure that proper Unicode support is worthwhile for Python 2 if it makes the code more complex. The Python 2 version still supports non-ASCII characters -- they just don't get decoded properly in e.g. help texts.

Just passing `encoding="utf-8"` for Python 3 would be another option.

@SebastianBoe How painful would it be to pass an UTF-8 locale setting to the Docker container? That's the most proper solution in a way, but I'm not opposed to implementing stuff that makes people's lives easier if the alternative is to have to muck around with stuff like that.

Once Python 3.7 is out, it should make things work automagically.

SebastianBoe:
"@SebastianBoe How painful would it be to pass an UTF-8 locale setting to the Docker container?"
I believe it is easy, this is 2018, if users have ascii as their system locale they should expect their non-ascii applications to break.

This change was intended to make things just-work, even for users have not configured their system properly. Or at the very least make kconfiglib behave consistently independent of the platform.

"Just passing encoding="utf-8" for Python 3 would be another option."
This would satisfy our requirements at least because we only use Python3.

@pfalcon: If I have understood this locale business correctly, neither Zephyr or Kconfiglib is explicitly supporting or not supporting UTF-8, instead they (reasonably?) use the system locale to decode text, so it is the public CI system (Docker image's system locale) that are failing to enable UTF-8. I may have misunderstood this though.

SebastianBoe:
@pfalcon : See this line in the first CI url that was posted:

02:02:24 + LANG=C
https://ci.linaro.org/job/lite-aeolus-jerryscript/PLATFORM=frdm_k64f,ZEPHYR_GCC_VARIANT=zephyr,label=docker-xenial-amd64-13/257/console

I believe this is setting C as the system locale, perhaps changing it to en_US.UTF-8 will resolve the CI issue.

ulfalizer:
@SebastianBoe 
Yup, Python just gets the default encoding from [the environment](http://pubs.opengroup.org/onlinepubs/7908799/xbd/envvar.html). Looks like setting `LANG` will work provided none of the `LC_*` variables are set. `LC_CTYPE` is the one that determines the encoding.

[PEP 538](https://www.python.org/dev/peps/pep-0538/) uses the following with Docker:

    LC_CTYPE=C.UTF-8

Not sure why the Docker people decided to default to ASCII. Seems to have caused a lot of pain.

ulfalizer:
@pfalcon @SebastianBoe 
Report back once you've tried the CI locale stuff. That might make other components in the container use UTF-8 too, which is nice.

Maybe a warning related to Docker could be added somewhere. I think I prefer getting the encoding from the environment for now at least, as there's some pain involved for Python 2 (plus it's nice to respect user preferences, though people probably often want UTF-8 even if it's not configured).

pfalcon:
@ulfalizer : Please assume that a user can't change "CI locale". Please assume that suddenly something started to break in somebody's builds because somebody else committed something. And the break is in the Kconfiglib project, and that's where fingers are pointed to.

I'd suggest to apply this patch, then watch if it leads to any issues (I'd estimate the probability is much lower than the issue fixed by the patch), then possibly add Kconfiglib option to let people specify non-standard encoding.

ulfalizer:
@pfalcon 
Unfortunately, the patch breaks Python 2.

Since all you need is Python 3 support (I think), an immediate fix would be to apply this patch to the Zephyr Kconfiglib branch. I've provided an extended patch at the end that also handles `string` symbol values with UTF-8 characters. It doesn't bother with `io.open()`, as it turns out that more than that would be needed for Python 2 anyway.

Adding the following to the beginning of `kconfig.py` might fix it as well, and would then also work for Python 2. It'd remove the need to patch `kconfiglib.py`.

    import locale
    locale.setlocale(locale.LC_CTYPE, "C.UTF-8")

How difficult would it be in your particular case to change the locale for the Docker containers? That would probably be a good idea in general, as other tools besides Python running inside the container will use it. The root cause of the problem is the locale setting in Docker. This is a symptom of that.

I'm not opposed in general to workarounds that make people's lives easier, even when the problem is "someone else's fault", but in this case it comes with tradeoffs related to code complexity and non-standard behavior. Before adding a workaround, I think we ought to check how hard it would be to fix the root cause of the problem.

If it turns out to be a lot more painful than just setting an environment variable, then I'd be happy to add a workaround. In the meantime, these quickfixes will work.

```
diff --git a/kconfiglib.py b/kconfiglib.py
index 99bdf80..98293b2 100644
--- a/kconfiglib.py
+++ b/kconfiglib.py
@@ -877,7 +877,7 @@ class Kconfig(object):
           would usually want it enclosed in '/* */' to make it a C comment,
           and include a final terminating newline.
         """
-        with open(filename, "w") as f:
+        with open(filename, "w", encoding="utf-8") as f:
             f.write(header)
 
             # Avoid duplicates -- see write_config()
@@ -938,7 +938,7 @@ class Kconfig(object):
           would usually want each line to start with '#' to make it a comment,
           and include a final terminating newline.
         """
-        with open(filename, "w") as f:
+        with open(filename, "w", encoding="utf-8") as f:
             f.write(header)
 
             # Symbol._written is set to True when a symbol config string is
@@ -1008,7 +1008,7 @@ class Kconfig(object):
           would usually want each line to start with '#' to make it a comment,
           and include a final terminating newline.
         """
-        with open(filename, "w") as f:
+        with open(filename, "w", encoding="utf-8") as f:
             f.write(header)
 
             # Avoid duplicates -- see write_config()
@@ -1170,7 +1170,7 @@ class Kconfig(object):
         # A separate helper function is neater than complicating write_config()
         # by passing a flag to it, plus we only need to look at symbols here.
 
-        with open("auto.conf", "w") as f:
+        with open("auto.conf", "w", encoding="utf-8") as f:
             for sym in self.defined_syms:
                 sym._written = False
 
@@ -1196,7 +1196,7 @@ class Kconfig(object):
             # No old values
             return
 
-        with open("auto.conf", _UNIVERSAL_NEWLINES_MODE) as f:
+        with open("auto.conf", _UNIVERSAL_NEWLINES_MODE, encoding="utf-8") as f:
             for line in f:
                 set_match = self._set_re_match(line)
                 if not set_match:
@@ -1340,12 +1340,12 @@ class Kconfig(object):
         # was set when the configuration was loaded
 
         try:
-            return open(filename, _UNIVERSAL_NEWLINES_MODE)
+            return open(filename, _UNIVERSAL_NEWLINES_MODE, encoding="utf-8")
         except IOError as e:
             if not os.path.isabs(filename) and self.srctree is not None:
                 filename = os.path.join(self.srctree, filename)
                 try:
-                    return open(filename, _UNIVERSAL_NEWLINES_MODE)
+                    return open(filename, _UNIVERSAL_NEWLINES_MODE, encoding="utf-8")
                 except IOError as e2:
                     # This is needed for Python 3, because e2 is deleted after
                     # the try block:
```

ulfalizer:
Are the containers managed centrally by the way? That might affect how easy it would be to fix.

ulfalizer:
I hacked up a patch that doesn't break Python 2. Python 2 never had an issue with invalid characters for the current encoding, as `str` isn't decoded in Python 2.

This is pretty ugly stuff. Can probably see why I want to avoid it. See the earlier discussion for why `encoding="utf8"` turns ugly for Python 2, even if `io.open()` is used.

```diff
diff --git a/kconfiglib.py b/kconfiglib.py
index 99bdf80..8fb8b7a 100644
--- a/kconfiglib.py
+++ b/kconfiglib.py
@@ -877,7 +877,7 @@ class Kconfig(object):
           would usually want it enclosed in '/* */' to make it a C comment,
           and include a final terminating newline.
         """
-        with open(filename, "w") as f:
+        with _open_utf8(filename, "w") as f:
             f.write(header)
 
             # Avoid duplicates -- see write_config()
@@ -938,7 +938,7 @@ class Kconfig(object):
           would usually want each line to start with '#' to make it a comment,
           and include a final terminating newline.
         """
-        with open(filename, "w") as f:
+        with _open_utf8(filename, "w") as f:
             f.write(header)
 
             # Symbol._written is set to True when a symbol config string is
@@ -1008,7 +1008,7 @@ class Kconfig(object):
           would usually want each line to start with '#' to make it a comment,
           and include a final terminating newline.
         """
-        with open(filename, "w") as f:
+        with _open_utf8(filename, "w") as f:
             f.write(header)
 
             # Avoid duplicates -- see write_config()
@@ -1170,7 +1170,7 @@ class Kconfig(object):
         # A separate helper function is neater than complicating write_config()
         # by passing a flag to it, plus we only need to look at symbols here.
 
-        with open("auto.conf", "w") as f:
+        with _open_utf8("auto.conf", "w") as f:
             for sym in self.defined_syms:
                 sym._written = False
 
@@ -1196,7 +1196,7 @@ class Kconfig(object):
             # No old values
             return
 
-        with open("auto.conf", _UNIVERSAL_NEWLINES_MODE) as f:
+        with _open_utf8("auto.conf", _UNIVERSAL_NEWLINES_MODE) as f:
             for line in f:
                 set_match = self._set_re_match(line)
                 if not set_match:
@@ -1340,12 +1340,12 @@ class Kconfig(object):
         # was set when the configuration was loaded
 
         try:
-            return open(filename, _UNIVERSAL_NEWLINES_MODE)
+            return _open_utf8(filename, _UNIVERSAL_NEWLINES_MODE)
         except IOError as e:
             if not os.path.isabs(filename) and self.srctree is not None:
                 filename = os.path.join(self.srctree, filename)
                 try:
-                    return open(filename, _UNIVERSAL_NEWLINES_MODE)
+                    return _open_utf8(filename, _UNIVERSAL_NEWLINES_MODE)
                 except IOError as e2:
                     # This is needed for Python 3, because e2 is deleted after
                     # the try block:
@@ -4281,6 +4281,11 @@ def _sym_to_num(sym):
     return sym.tri_value if sym.orig_type in (BOOL, TRISTATE) else \
            int(sym.str_value, _TYPE_TO_BASE[sym.orig_type])
 
+def _open_utf8_py3(filename, mode):
+    return open(filename, mode, encoding="utf-8")
+
+_open_utf8 = open if sys.version_info[0] < 3 else _open_utf8_py3
+
 def _stderr_msg(msg, filename, linenr):
     if filename is not None:
         msg = "{}:{}: {}".format(filename, linenr, msg)
```

ulfalizer:
@pfalcon 
I'm a Docker noob, but this seems to the common fix from some googling. Docker and UTF-8 seems to be pretty shitstormy. It's bitten a bunch of Java and SQL people too.

```
ENV LANG C.UTF-8
```

Almost get the feeling they won't make it the default out of spite at this point...

pfalcon:
> @pfalcon
> Unfortunately, the patch breaks Python 2.

Ah, indeed, I missed that "basestring" issue. Thanks for suggestion how to fix issue I described in https://github.com/ulfalizer/Kconfiglib/pull/41#issuecomment-372801002 on our side. As I mentioned, it was worked around on Zephyr's side so far. As for "LANG=C", it wasn't added by me, so I'm not keen to change it. If nothing else, it does show the problem with all this stuff - that behavior of the system (Zephyr build system in our case) is not stable and depends on external factors. Fairly speaking, that's what CI is for - to detect problems ahead of them hitting human users (which are notoriously bad at reporting issues, it's often "your stuff doesn't build"). Of course, I'll have to make locale changes if the problem reoccurs.

ulfalizer:
@pfalcon 
Maybe you could email whoever is in charge of the containers and ask them if they could change it. You could tell them to google "docker utf-8" as well, as that turns up other problems. Other components are affected too, not just Python.

The problem seems to mostly be constrained to Docker, as all the mainstream distributions already default to UTF-8. This is the first time I've run across it with Kconfiglib.

I can't really speak for the motivations of the Docker people, as I haven't even used it, but to me it looks like a bad default. It breaks Unicode support for tools that set their encoding from the environment (which is probably most of them).

SebastianBoe:
Zephyr only has one script that imports kconfiglib.py, so doing this seems like the simplest solution to me:

> Adding the following to the beginning of kconfig.py might fix it as well, and would then also work for Python 2. It'd remove the need to patch kconfiglib.py.
> 
> import locale
> locale.setlocale(locale.LC_CTYPE, "C.UTF-8")

I'll close this PR, it is replaced by the Zephyr PR: https://github.com/zephyrproject-rtos/zephyr/pull/6716

Thank you for discovering this technique, I did not think of this approach at all.

This means that;

- Zephyr doesn't depend on how the system locale is configured
- Zephyr doesn't have to maintain a fork
- Kconfiglib.py continues to support Py2 well and rely on the system locale

PR #42, opened 2018-03-18
Add print_config_tree.py example script

rfrail3:
I was looking for a solution to get the full hierachical structure of the kernel configuration, similar to the "make menuconfig" but with all the menus expanded. I asked here [1] and a user pointed me to your proyect.

The script print_config_tree.py prints the menu entries as a tree with its value in the .config file. It is done taking your menuconfig.py file as starting point.

With it and a diff tool, is easy to compare, side by side, different kernel configurations and understand it.

I also add a switch to print the help too, it can be helpful, but it's disabled by default because the output is too large.

[1] - https://unix.stackexchange.com/questions/430693/formatted-print-of-linux-kernel-config

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -238,6 +238,8 @@ The `examples/ <https://github.com/ulfalizer/Kconfiglib/blob/master/examples>`_ 
 
 - `print_tree.py <https://github.com/ulfalizer/Kconfiglib/blob/master/examples/print_tree.py>`_ prints a tree of all configuration items.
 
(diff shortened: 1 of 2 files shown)


ulfalizer:
Hello,

Neat idea. I'm happy to take it.

I did some small style consistency cleanups and removed a few unused things originating from `menuconfig.py`. See [this commit](https://github.com/ulfalizer/Kconfiglib/commit/98574bfcf3ad629f6fd5eb5cd9824dae75d9e2bf).


If you're happy with it, push a new squashed version as a single commit, and I'll merge it in.

Cheers,
Ulf

ulfalizer:
Hmm... looks like i typo'd a bit: The second 'between' in "...handy e.g. for diffing between different between .config files..." shouldn't be there. Could you fix that at the same time? Thanks.

rfrail3:
Thanks for update and clean it. Here are all the changes in one commit, I hope that it's all ok.
Regards,

ulfalizer:
Looks fine. Merged!

Thanks!

PR #50, opened 2018-08-21
Add a generic node iterator.

pinkfluid:
Use the node iterator to traverse symbols/nodes in write_config(),
write_min_config(), write_autoconfs()...

This ensures that the order of symbols is consistent between config
files and headers and that symbols don't get shuffled around between
configuration changes.

A nice side effect is that comments are now included in the generated
header files and minimal configs.

I did run the test suite against linux kernel 4.18.3 and it seems to be failing, but not because of this change. Please let me know if you need the logs.


kconfiglib.py
```diff
@@ -995,6 +995,50 @@ def _load_config(self, filename, replace):
                 if not choice._was_set:
                     choice.unset_value()
 
+    def _node_items(self):
```
ulfalizer: Might be nice to have this available in the public API as well.

What do you think of calling the function `node_iter(skip_duplicates=False)`? `skip_duplicates=True` would use the existing logic, while `skip_duplicates=False` would return all nodes.

pinkfluid:
Hi ulfalizer,

if you have already a patch that guarantees the order of symbols in "min_config" and header files that's great :) 

The issue I was trying to fix is that we're having generated header files and "minimal" configs under revision control and the symbols were shuffled around randomly when the Kconfig changed.

Anyway, I did push changes that should satisfy all if not most of your requests. That being said, I'll be happy with only the ordering patch as that should be enough to fix the issues we're having.

Thanks for thins great library!

ulfalizer:
@pinkfluid 
Yep... finishing up a fix now (it's basically just replacing `for sym in self._defined_syms_set:` with `for sym in self.defined_syms`, but I want to add some tests as well).

Some kind of tree walking helper would definitely be nice to have still, and having the comments in the headers might be nice too. I'll look into a bit later.

pinkfluid:
That's great, thank you.

pinkfluid:
Just one more comment, node_iter() can probably be changed to use the "defined_syms" when skip_duplicates=False.

ulfalizer:
@pinkfluid 
Might want to visit other `MenuNode`s besides symbols though.

A `MenuNode` can represent a symbol, a choice, a menu, or a comment. `skip_duplicates` would only apply to symbols, because it's usually what you want when writing output.

ulfalizer:
Pushed [a fix](https://github.com/ulfalizer/Kconfiglib/commit/d78c293c41351256f76fb2fa27e72e399b130e4f) for the ordering issue and made a new 10.0.1 release.

I noticed that `write_config()` could be simplified a bit too: https://github.com/ulfalizer/Kconfiglib/commit/db6cb68b74bc3a9f88f802a0f32613f95df7b854. Maybe that could be integrated here as well.

Gotta work on some work-related things now, but I'll get back to the traversal helper later. :)

pinkfluid:
Hi, the new version seems to work perfectly. The order seems to be preserved in both minconfig an config.h! Thank you!

ulfalizer:
I added a variant of the node iterator now as well: https://github.com/ulfalizer/Kconfiglib/commit/7dae98803a6fc5d08041d1387e2e0d83fc0eb0ed

I did some reorganization prior, and it's a bit hard to write docs without knowing all the Kconfig details, so didn't want to force you to update the PR.

pinkfluid:
I was already happy with the ordering fix, but this looks great :)

I didn't really take a look deep inside the node class, but could this evolve to some sort of a public API?

It's just a thought, but you could use the iterator to generate customized files. For example, some projects may need to  add support for other languages besides C or generate a C structure instead of a header file... 

I think its time we discard this pull request :)

ulfalizer:
@pinkfluid 
The `MenuNode` class and the new `Kconfig.node_iter()` generator are already public APIs. You can access the top menu node via `Kconfig.top_node` and follow `next` and `list` pointers, etc., as well. Symbols and choices also have a public `nodes` property for accessing their menu nodes.

See the [print_tree.py](https://github.com/ulfalizer/Kconfiglib/blob/master/examples/print_tree.py) example for example.

Alright, closing the issue. Thanks!

ulfalizer:
Basically, anything that doesn't have a `_` prefix is a public API. Public functions have docstrings too, while internal functions use "doc comments" instead.

pinkfluid:
When I said "public API", I was thinking more in the lines of creating an API for writing configurations using a generator. Something you would use instead of write_config()/write_min_config(). This could be useful, for example, for export the configuration as a JSON or INI files, and probably other things like:

- adding support for non-C/shell/make languages
- kconfiglib could be used for generating configuration files (imagine having a menuconfig interface for configuring a web server?)
- writing to stdout/sockets

These are just ideas, so its up to you to decide if its useful or not :) In any case, I'm rather happy with kconfiglib in its current state.

This is an example of such iterator that I scrapped together. Below you can find examples that export the current config in JSON and INI formats. 

```python
import kconfiglib

class WriteItem:
    MENU = 1
    COMMENT = 2
    BOOL = 3
    TRISTATE = 4
    STRING = 5
    INT = 6
    HEX = 7

    def __init__(self, wtype, name, value):
        self.type = wtype
        self.name = name
        self.value = value


def write_items(self, unique_syms=False, minimal=False):
    for node in self.node_iter(unique_syms):
        item = node.item

        if isinstance(item, kconfiglib.Symbol):
            if minimal:
                # Skip symbols that cannot be changed. Only check
                # non-choice symbols, as selects don't affect choice
                # symbols.
                if not item.choice and \
                   item.visibility <= kconfiglib.expr_value(item.rev_dep):
                    continueFor example

                # Skip symbols whose value matches their default
                if item.str_value == item._str_default():
                    continue

                # Skip symbols that would be selected by default in a
                # choice, unless the choice is optional or the symbol type
                # isn't bool (it might be possible to set the choice mode
                # to n or the symbol to m in those cases).
                if item.choice and \
                   not item.choice.is_optional and \
                   item.choice._get_selection_from_defaults() is node and \
                   item.orig_type == BOOL and \
                   item.tri_value == 2:
                    continue


            item_name = "{}{}".format(self.config_prefix, item.name)
            if item.orig_type == kconfiglib.BOOL:
                yield WriteItem(WriteItem.BOOL, item_name, item.str_value == "y")

            elif item.orig_type == kconfiglib.TRISTATE:
                if item.str_value == "m":
                    yield WriteItem(
                            WriteItem.TRISTATE,
                            "{}_MODULE".format(item_name),
                            True)
                else:
                    yield WriteItem(
                            WriteItem.TRISTATE,
                            item_name,
                            item.str_value == "y")

            elif item.orig_type == kconfiglib.STRING:
                yield WriteItem(WriteItem.STRING, item_name, item.str_value)
            elif item.orig_type == kconfiglib.INT:
                yield WriteItem(WriteItem.INT, item_name, int(item.str_value) if item.str_value else 0)
            elif item.orig_type == kconfiglib.HEX:
                yield WriteItem(WriteItem.HEX, item_name, int(item.str_value, 16) if item.str_value else 0)
            else:
                continue
        elif not kconfiglib.expr_value(node.dep):
            pass
        elif item == kconfiglib.COMMENT:
            if not kconfiglib.expr_value(node.dep): continue
            yield WriteItem(WriteItem.COMMENT, None, node.prompt[0])
        elif item == kconfiglib.MENU and kconfiglib.expr_value(node.visibility):
            if not kconfiglib.expr_value(node.dep): continue
            yield WriteItem(WriteItem.MENU, None, node.prompt[0])
        else:
            # Skip everything else
            pass

kconf = kconfiglib.Kconfig("/home/mitja/git/device/core/kconfig/Kconfig")
kconf.load_config("/home/mitja/git/device/vendor/plume/kconfig/targets/VFE_BCM52")


"""
Dump config in JSON format
"""
dot = False

print("{")
for w in write_items(kconf, minimal=False):
    if dot: print(",\n", end="")
    dot = True

    if w.type in (WriteItem.HEX, WriteItem.INT):
        print('     "%s": %d' % (w.name, w.value), end='')
    elif w.type == WriteItem.STRING:
        print('     "%s": "%s"' % (w.name, w.value), end='')
    elif w.type in (WriteItem.BOOL, WriteItem.TRISTATE):
        print('     "%s": %s' % (w.name, "true" if w.value else "false"), end='')
    elif w.type in (WriteItem.COMMENT, WriteItem.MENU):
        print('\n     "__COMMENT__": "{}"'.format(w.value), end='')
    else:
        dot = False

print("\n}")


"""
Generate an INI file
"""
kconf.config_prefix = ""
for w in write_items(kconf, minimal=False):
    if w.type in (WriteItem.HEX, WriteItem.INT):
        print('%s=%d' % (w.name, w.value))
    elif w.type == WriteItem.STRING:
        print('%s="%s"' % (w.name, w.value))
    elif w.type in (WriteItem.BOOL, WriteItem.TRISTATE):
        print('%s=%s' % (w.name, "true" if w.value else "false"))
    elif w.type in (WriteItem.COMMENT, WriteItem.MENU):
        print('\n\n[{}]'.format(w.value))
```

ulfalizer:
The tricky part is coming up with something general I think. Always the risk that you make something too limited if you make it too magic.

I think I prefer the approach of adding small helpers whenever you find that you keep doing the same thing over and over. Just gotta discover those helpers.

pinkfluid:
It was just an idea, I don't have any immediate use of it right now :) I'll just push it to my fork, so I have a backup in case I need this in the future.

Thanks again.

PR #53, opened 2018-08-29
menuconfig: Add support for custom color schemes (aka styles)

pinkfluid:
I've added experimental basic themeing (styling) support to menuconfig.

The code is probably not the greatest, but might be a source of few ideas. I'm sending a pull request just so its easier to view the diff. Please close it at your leisure.

There are two sources of styles:
- Kconfig files by using the '# $style: ....' keyword
- The MENUCONFIG_STYLE environment variable

By default menuconfig starts off with the default theme. The default
theme can be changed by using the following keywords:

        path, separator, list, selection, inv_list, inv_selection, help,
        frame, body, edit, jump_to_edit, text

Colors can be define symbolically, for example: "bold_bright_yellow_on_blue"

or using the 256-color scheme:

"underline_color226_on_color33"

A special keyword "style=" is used to use a builtin theme instead of the
default.

Examples:

Use aquatic theme:

MENUCONFIG_STYLE="style=aquatic"

Use a theme with a purple accent instead of yellow:

MENUCONFIG_STYLE="selection=underline_bold_white_on_purple separator=white_on_purple"

Aquatic theme with purple accent:

MENUCONFIG_STYLE="style=aquatic selection=underline_bold_white_on_purple separator=white_on_purple"

Kconfig example of the above:

        # $style: style=aquatic
        # $style: selection=underline_bold_white_on_purple
        # $style: separator=white_on_purple

The style is parsed from all soruce Kconfig files, so the theme can
reside in its own file and can be sourced at the beginning.

In the future, support for JSON files are something similar can be
added.

### DIFF
--- a/menuconfig.py
+++ b/menuconfig.py
@@ -48,12 +48,68 @@ $srctree is supported through Kconfiglib.
 Color schemes
 =============
 
-Setting the environment variable MENUCONFIG_THEME to 'aquatic' will enable an
+It is possible to customize the color scheme by setting the MENUCONFIG_STYLE
+environment variable. For example, setting it to 'aquatic' will enable an
 alternative, less yellow, more 'make menuconfig'-like color scheme, contributed
 by Mitja Horvat (pinkfluid).
 
-See the _init_styles() function if you want to add additional themes. I'm happy
-to take them in upstream.
+This is the current list of built-in styles:
+    - default       classic Kconfiglib theme with a yellow accent
+    - monochrome    colorless theme (uses only bold and standout) attributes,
+                    this style is used if the terminal doesn't support colors
+    - aquatic       blue tinted style loosely resembling the lxdialog theme
+
+It is possible to customize the current style by changing colors of UI
+elements on the screen. This is the list of elements that can be stylized:
+
+    - path          Top row in the main display, with the menu path
+    - separator     Separator lines between windows. Also used for the top line
+                    in the symbol
+    - list          List of items, e.g. the main display
+    - selection     Style for the selected item
+    - inv-list:     Like list, but for invisible items. Used in show-all mode.
+    - inv-selection Like selection, but for invisible items. Used in show-all
+                    mode.
+    - help          Help text windows at the bottom of various fullscreen
+                    dialogs
+    - frame         Frame around dialog boxes
+    - body          Body of dialog boxes
+    - edit          Edit box in pop-up dialogs
+    - jump-edit     Edit box in jump-to dialog
+    - text          Symbol information text
+
+The color definition is a comma separated list of attributes:
+
+    - fg:COLOR      Set the foreground/background colors. COLOR can be one of
+      * or *        the basic 16 colors (black, red, green, yellow, blue,
+    - bg:COLOR      magenta,cyan, white and brighter versions, for example,
+                    brightred). On terminals that support 256 color mode, you
+                    can use the colorNN keyword, where NN is a number between 1
+                    and 255.
+
+                    Note: On some terminals a bright version of the color implies
+                    bold.
+    - bold          Use bold text
+    - underline     Use underline text
+    - standout      Standout text attribute (reverse color)
+
+A keyword without the '=' is assumed to be a style template. The template name
+is looked up in the built-in styles list and the style definition is expanded
+in-place. With this, built-in styles can be used as basis for new styles.
+
+For example, take the aquatic theme and give it a red selection bar:
+
+MENUCONFIG_STYLE="aquatic selection=fg:white,bg:red"
+
+If there's an error in the style definition, a StyleError exception will be
+thrown briefly describing the error.
+
+The 'default' theme is always implicitly parsed first (or the 'monochrome'
+theme if the terminal lacks colors), so the following two settings have the
+same effect:
+
+    MENUCONFIG_STYLE="selection=fg:white,bg:red"
+    MENUCONFIG_STYLE="default selection=fg:white,bg:red"
 
 
 Other features
@@ -165,101 +221,172 @@ strings/regexes to find entries that match all of them. Type Ctrl-F to
 view the help of the selected item without leaving the dialog.
 """[1:-1].split("\n")
 
+_STYLES = {
+    "default": """
+    path=fg:black,bg:white,bold
+    separator=fg:black,bg:yellow,bold
+    list=fg:black,bg:white
+    selection=fg:white,bg:blue,bold
+    inv-list=fg:red,bg:white
+    inv-selection=fg:red,bg:blue
+    help=fg:black,bg:white,bold
+    frame=fg:black,bg:yellow,bold
+    body=fg:white,bg:black
+    edit=fg:white,bg:blue
+    jump-edit=fg:white,bg:blue
+    text=fg:black,bg:white
+    """,
+
+    # This style is forced on terminals that do no support colors
+    "monochrome": """
+    path=bold
+    separator=bold,standout
+    list=
+    selection=bold,standout
+    inv-list=bold
+    inv-selection=bold,standout
+    help=bold
+    frame=bold,standout
+    body=
+    edit=standout
+    jump-edit=
+    text=
+    """,
+
+    # Blue tinted style loosely resembling lxdialog
+    "aquatic": """
+    path=fg:cyan,bg:blue,bold
+    separator=fg:white,bg:cyan,bold
+    help=fg:cyan,bg:blue,bold
+    frame=fg:white,bg:cyan,bold
+    body=fg:brightwhite,bg:blue
+    edit=fg:black,bg:white
+    """
+}
+
+# Standard colors definition
+_STYLE_STD_COLORS = {
+    # Basic colors
+    "black":        curses.COLOR_BLACK,
+    "red":          curses.COLOR_RED,
+    "green":        curses.COLOR_GREEN,
+    "yellow":       curses.COLOR_YELLOW,
+    "blue":         curses.COLOR_BLUE,
+    "magenta":      curses.COLOR_MAGENTA,
+    "cyan":         curses.COLOR_CYAN,
+    "white":        curses.COLOR_WHITE,
+
+    # Bright versions
+    "brightblack":  curses.COLOR_BLACK + 8,
+    "brightred":    curses.COLOR_RED + 8,
+    "brightgreen":  curses.COLOR_GREEN + 8,
+    "brightyellow": curses.COLOR_YELLOW + 8,
+    "brightblue":   curses.COLOR_BLUE + 8,
+    "brightmagenta":curses.COLOR_MAGENTA + 8,
+    "brightcyan":   curses.COLOR_CYAN + 8,
+    "brightwhite":  curses.COLOR_WHITE + 8,
+
+    # Aliases
+    "purple":       curses.COLOR_MAGENTA,
+    "brightpurple": curses.COLOR_MAGENTA + 8,
+    "default":  -1
+}
+
+_style = {}
+
+class StyleError(Exception):
+    pass
+
+def _parse_style(style_str):
+    global _style
+
+    for sline in style_str.split():
+        # Words without a "=" character represents a style template
+        if "=" in sline:
+            key, data = sline.split("=")
+            _style[key] = _style_to_curses(data)
+        else:
+            # Recursively expand style templates
+            if sline not in _STYLES:
+                raise StyleError("Unknown built-in style: {}".format(sline))
+
+            # Merge in the template
+            _parse_style(_STYLES[sline])
+
+def _style_to_curses(cstr):
+    """
+    Parse a style definition and convert it to curses attributes
+
+    This function returns a list of: (fg_color, bg_color, attributes)
+    """
+    def parse_color(t):
+        ts = t.split(":")
+        if len(ts) != 2:
+            raise StyleError("Invalid color notation: {}".format(t))
+
+        cdef = ts[1]
+
+        if cdef in _STYLE_STD_COLORS:
+            return _STYLE_STD_COLORS[cdef]
+
+        if cdef.startswith("color"):
+            cnum = cdef[len("color"):]
+            if not cnum.isdigit():
+                raise StyleError(
+                        "Fixed color must be followed by a number. " \
+                        "Error near: {}".format(t))
+
+            cnum = int(cnum)
+            if not -1 <= cnum <= curses.COLORS:
+                raise StyleError(
+                        "Fixed color must be in range 0..{}. " \
+                        "Error near: {}".format(curses.COLORS - 1, t))
+            return cnum
+
+        raise StyleError("Unknown color definition: {}".format(t))
+
+    attrs = 0
+    fg_color = -1
+    bg_color = -1
+
+    if not cstr:
+        return _style_attr(fg_color, bg_color, attrs)
+
+    # Parse attributes
+    for t in cstr.split(","):
+        if t == "bold":
+            # A_BOLD tends to produce faint and hard-to-read text on the Windows
+            # console, especially with the old color scheme, before the
+            # introduction of
+            # https://blogs.msdn.microsoft.com/commandline/2017/08/02/updating-the-windows-console-colors/
+            attrs |= curses.A_NORMAL if _IS_WINDOWS else curses.A_BOLD
+        elif t == "standout":
+            attrs |= curses.A_STANDOUT
+        elif t == "underline":
+            attrs |= curses.A_UNDERLINE
+        elif t.startswith("fg:"):
+            fg_color = parse_color(t)
+        elif t.startswith("bg:"):
+            bg_color = parse_color(t)
+        else:
+            raise StyleError("Invalid style attribute: {}".format(cstr))
+
+    return _style_attr(fg_color, bg_color, attrs)
+
 def _init_styles():
-    global _PATH_STYLE
-    global _SEPARATOR_STYLE
-    global _LIST_STYLE
-    global _LIST_SEL_STYLE
-    global _LIST_INVISIBLE_STYLE
-    global _LIST_INVISIBLE_SEL_STYLE
-    global _HELP_STYLE
-    global _DIALOG_FRAME_STYLE
-    global _DIALOG_BODY_STYLE
-    global _DIALOG_EDIT_STYLE
-    global _JUMP_TO_EDIT_STYLE
-    global _INFO_TEXT_STYLE
+    if curses.has_colors():
+        curses.use_default_colors()
 
-    # Initialize styles for different parts of the application. The arguments
-    # are ordered as follows:
-    #
-    #   1. Text color
-    #   2. Background color
-    #   3. Attributes
-    #   4. Extra attributes if colors aren't available. The colors will be
-    #      ignored in this case, and the attributes from (3.) and (4.) will be
-    #      ORed together.
+    # Force the monochrome style
+    if not curses.has_colors():
+        _parse_style("monochrome")
+    # Use the default style
+    else:
+        _parse_style("default")
 
-    # A_BOLD tends to produce faint and hard-to-read text on the Windows
-    # console, especially with the old color scheme, before the introduction of
-    # https://blogs.msdn.microsoft.com/commandline/2017/08/02/updating-the-windows-console-colors/
-    BOLD = curses.A_NORMAL if _IS_WINDOWS else curses.A_BOLD
-
-    # Default styling. Themes can override these settings below.
-
-    # Top row in the main display, with the menu path
-    PATH_STYLE               = (curses.COLOR_BLACK, curses.COLOR_WHITE,  BOLD                              )
-
-    # Separator lines between windows. Also used for the top line in the symbol
-    # information dialog.
-    SEPARATOR_STYLE          = (curses.COLOR_BLACK, curses.COLOR_YELLOW, BOLD,            curses.A_STANDOUT)
-
-    # List of items, e.g. the main display
-    LIST_STYLE               = (curses.COLOR_BLACK, curses.COLOR_WHITE,  curses.A_NORMAL                   )
-
-    # Style for the selected item
-    LIST_SEL_STYLE           = (curses.COLOR_WHITE, curses.COLOR_BLUE,   BOLD,            curses.A_STANDOUT)
-
-    # Like _LIST_(SEL_)STYLE, for invisible items. Used in show-all mode.
-    LIST_INVISIBLE_STYLE     = (curses.COLOR_RED,   curses.COLOR_WHITE,  curses.A_NORMAL, BOLD             )
-    LIST_INVISIBLE_SEL_STYLE = (curses.COLOR_RED,   curses.COLOR_BLUE,   BOLD,            curses.A_STANDOUT)
-
-    # Help text windows at the bottom of various fullscreen dialogs
-    HELP_STYLE               = PATH_STYLE
-
-    # Frame around dialog boxes
-    DIALOG_FRAME_STYLE       = SEPARATOR_STYLE
-
-    # Body of dialog boxes
-    DIALOG_BODY_STYLE        = (curses.COLOR_WHITE, curses.COLOR_BLACK,  curses.A_NORMAL                   )
-
-    # Edit box in pop-up dialogs
-    DIALOG_EDIT_STYLE        = (curses.COLOR_WHITE, curses.COLOR_BLUE,   curses.A_NORMAL, curses.A_STANDOUT)
-
-    # Edit box in jump-to dialog
-    JUMP_TO_EDIT_STYLE       = (curses.COLOR_WHITE, curses.COLOR_BLUE,   curses.A_NORMAL,                  )
-
-    # Symbol information text
-    INFO_TEXT_STYLE          = LIST_STYLE
-
-    if os.environ.get("MENUCONFIG_THEME") == "aquatic":
-        # More 'make menuconfig'-like theme, contributed by Mitja Horvat
-        # (pinkfluid)
-        PATH_STYLE         = (curses.COLOR_CYAN,  curses.COLOR_BLUE,  BOLD                              )
-        SEPARATOR_STYLE    = (curses.COLOR_WHITE, curses.COLOR_CYAN,  BOLD,            curses.A_STANDOUT)
-        HELP_STYLE         = PATH_STYLE
-        DIALOG_FRAME_STYLE = SEPARATOR_STYLE
-        DIALOG_BODY_STYLE  = (curses.COLOR_WHITE, curses.COLOR_BLUE,  curses.A_NORMAL                   )
-        DIALOG_EDIT_STYLE  = (curses.COLOR_BLACK, curses.COLOR_WHITE, curses.A_NORMAL, curses.A_STANDOUT)
-
-    # Turn styles into attributes and store them in global variables. Doing
-    # this separately minimizes the number of curses color pairs, and shortens
-    # the style definitions a bit.
-    #
-    # Could do some locals()/globals() trickery here too, but keep it
-    # searchable.
-    _PATH_STYLE               = _style(*PATH_STYLE)
-    _SEPARATOR_STYLE          = _style(*SEPARATOR_STYLE)
-    _LIST_STYLE               = _style(*LIST_STYLE)
-    _LIST_SEL_STYLE           = _style(*LIST_SEL_STYLE)
-    _LIST_INVISIBLE_STYLE     = _style(*LIST_INVISIBLE_STYLE)
-    _LIST_INVISIBLE_SEL_STYLE = _style(*LIST_INVISIBLE_SEL_STYLE)
-    _HELP_STYLE               = _style(*HELP_STYLE)
-    _DIALOG_FRAME_STYLE       = _style(*DIALOG_FRAME_STYLE)
-    _DIALOG_BODY_STYLE        = _style(*DIALOG_BODY_STYLE)
-    _DIALOG_EDIT_STYLE        = _style(*DIALOG_EDIT_STYLE)
-    _JUMP_TO_EDIT_STYLE       = _style(*JUMP_TO_EDIT_STYLE)
-    _INFO_TEXT_STYLE          = _style(*INFO_TEXT_STYLE)
-
+    # Use the user-defined style from the environemnt
+    if "MENUCONFIG_STYLE" in os.environ:
+        _parse_style(os.environ["MENUCONFIG_STYLE"])
 
 #
 # Main application
@@ -270,16 +397,15 @@ def _init_styles():
 #
 # Obscure Python: We never pass a value for color_attribs, and it keeps
 # pointing to the same dict. This avoids a global.
-def _style(fg_color, bg_color, attribs, no_color_extra_attribs=0,
-           color_attribs={}):
+def _style_attr(fg_color, bg_color, attribs, color_attribs={}):
     # Returns an attribute with the specified foreground and background color
     # and the attributes in 'attribs'. Reuses color pairs already created if
     # possible, and creates a new color pair otherwise.
     #
-    # Returns 'attribs | no_color_extra_attribs' if colors aren't supported.
+    # Returns 'attribs' if colors aren't supported.
 
     if not curses.has_colors():
-        return attribs | no_color_extra_attribs
+        return attribs
 
     if (fg_color, bg_color) not in color_attribs:
         # Create new color pair. Color pair number 0 is hardcoded and cannot be
@@ -333,7 +459,6 @@ def menuconfig(kconf):
     global _conf_changed
 
     _kconf = kconf
-
 
     _config_filename = standard_config_filename()
 
@@ -623,20 +748,20 @@ def _init():
     # Initialize windows
 
     # Top row, with menu path
-    _path_win = _styled_win(_PATH_STYLE)
+    _path_win = _styled_win(_style["path"])
 
     # Separator below menu path, with title and arrows pointing up
-    _top_sep_win = _styled_win(_SEPARATOR_STYLE)
+    _top_sep_win = _styled_win(_style["separator"])
 
     # List of menu entries with symbols, etc.
-    _menu_win = _styled_win(_LIST_STYLE)
+    _menu_win = _styled_win(_style["list"])
     _menu_win.keypad(True)
 
     # Row below menu list, with arrows pointing down
-    _bot_sep_win = _styled_win(_SEPARATOR_STYLE)
+    _bot_sep_win = _styled_win(_style["separator"])
 
     # Help window with keys at the bottom
-    _help_win = _styled_win(_HELP_STYLE)
+    _help_win = _styled_win(_style["help"])
 
     # The rows we'd like the nodes in the parent menus to appear on. This
     # prevents the scroll from jumping around when going in and out of menus.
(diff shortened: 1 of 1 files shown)


menuconfig.py
```diff
@@ -165,100 +165,276 @@
...
+    for kfile in _kconf.kconfig_filenames:
+        with open(kfile) as kf:
+            for kl in kf.readlines():
+                kl = kl.rstrip("\n")
+                #m = re.match("^\s*#\s+$style:\s+(a-z0-9=_\s)+$", kl)
+                m = re.match("^\s*#\s+\$style:\s*([a-z0-9=_\s]+)$", kl)
+                if m:
+                    kstyle += m[1] + " "
+
+    return style_from_string(kstyle)
```
ulfalizer: Not sure it's worthwhile to parse style settings from Kconfig files. Styling feels like something that should be up to the user, not to the Kconfig authors.

I think it would make more sense to just look at the top-level Kconfig file too. A single menu can span multiple Kconfig files, so it wouldn't make sense to have per-Kconfig-file settings either. Parsing through all Kconfig files is pretty expensive too (at least 0.17 seconds with Linux 4.18 on my machine).

I'll take a look at the rest of the styling code a bit later.

ulfalizer:
Instead of having `style=...`, maybe anything that doesn't have a `=` in it could be taken as a predefined style, and work the same as inserting all the settings from the style at that location.

You'd then be able to write stuff like `MENUCONFIG_STYLE=aquatic` and `MENUCONFIG_STYLE='aquatic separator=white_on_purple'`, where the later one overrides a single setting.

The default style would be implicit, so the later example would work the same as `MENUCONFIG_STYLE='default aquatic separator=white_on_purple'`.

Maybe setting could be separated by `,`, like this:

    MENUCONFIG_STYLE='aquatic separator=blue_on_red,bold selection=white_on_yellow,standout'

Maybe there could be separate `fg:<color>` and `bg:<color>` attributes too, instead of having specialized parsing for colors:

    MENUCONFIG_STYLE='aquatic separator=fg:blue,bg:red,bold selection=...'

Does `__` do anything special at the module level by the way? Only seen it within classes.

menuconfig.py
```diff
@@ -165,100 +165,276 @@
...
-    global _LIST_INVISIBLE_STYLE
-    global _LIST_INVISIBLE_SEL_STYLE
-    global _HELP_STYLE
-    global _DIALOG_FRAME_STYLE
-    global _DIALOG_BODY_STYLE
-    global _DIALOG_EDIT_STYLE
-    global _JUMP_TO_EDIT_STYLE
-    global _INFO_TEXT_STYLE
+# Default menuconfig theme
+DEFAULT_STYLE = \
```
ulfalizer: Should have a single initial `_`, to mark it as internal. That hides it in generated documentation, and when people do `from menuconfig import *` (though that's unlikely).

The only public interface in this module so far is the `menuconfig()` function, so all other global variables/functions get a leading `_`. Local variables never get a leading `_` though.

menuconfig.py
```diff
@@ -165,100 +165,276 @@
...
+        # Frame around dialog boxes
+        "frame":            "bold black on yellow",
+        # Body of dialog boxes
+        "body":             "white on black",
+        # Edit box in pop-up dialogs
+        "edit":             "white on blue",
+        # Edit box in jump-to dialog
+        "jump_to_edit":     "white on blue",
+        # Symbol information text
+        "text":             "black on white"
```
ulfalizer: Instead of having `_DEFAULT_STYLE` be a dictionary, I wonder if it could just be a string, in the normal `MENUCONFIG_STYLE` format. Everything would then go through a single style parsing code path.

The only difference for the default style would be that it's hardcoded to be parsed before `MENUCONFIG_STYLE`. This could be done by just adding `default` or the like to the beginning of the `MENUCONFIG_STYLE` string in the code, before parsing it.

menuconfig.py
```diff
@@ -165,100 +165,276 @@
...
+        "selection":        "bold standout",
+        "inv_list":         "bold",
+        "inv_selection":    "bold standout",
+        "help":             "bold",
+        "frame":            "bold standout",
+        "body":             "",
+        "edit":             "standout",
+        "jump_to_edit":     "",
+        "text":             ""
+}
```
ulfalizer: Ditto for this style.

menuconfig.py
```diff
@@ -165,100 +165,276 @@
...
+AQUATIC_STYLE = \
+{
+        "style":            "default",
+        "path":             "bold cyan on blue",
+        "separator":        "bold white on cyan",
+        "help":             "bold cyan on blue",
+        "frame":            "bold white on cyan",
+        "body":             "white on blue",
+        "edit":             "black on white",
+}
```
ulfalizer: Ditto for this style.

menuconfig.py
```diff
@@ -165,100 +165,276 @@
...
+
+__STYLES = {}
```
ulfalizer: Just a single leading `_`.

menuconfig.py
```diff
@@ -165,100 +165,276 @@
...
-    if os.environ.get("MENUCONFIG_THEME") == "aquatic":
-        # More 'make menuconfig'-like theme, contributed by Mitja Horvat
-        # (pinkfluid)
-        PATH_STYLE         = (curses.COLOR_CYAN,  curses.COLOR_BLUE,  BOLD                              )
-        SEPARATOR_STYLE    = (curses.COLOR_WHITE, curses.COLOR_CYAN,  BOLD,            curses.A_STANDOUT)
-        HELP_STYLE         = PATH_STYLE
-        DIALOG_FRAME_STYLE = SEPARATOR_STYLE
-        DIALOG_BODY_STYLE  = (curses.COLOR_WHITE, curses.COLOR_BLUE,  curses.A_NORMAL                   )
-        DIALOG_EDIT_STYLE  = (curses.COLOR_BLACK, curses.COLOR_WHITE, curses.A_NORMAL, curses.A_STANDOUT)
+    def __parse_main(tok):
```
ulfalizer: Just a single leading `_`.

menuconfig.py
```diff
@@ -165,100 +165,276 @@
...
+    }
+
+    def __parse_tokenizer(cdef):
+        for cs in cdef.replace(' ', '_').split('_'):
+            if cs == '': continue
+            yield cs
+
+        yield None
+
+    def __parse_color(tok, t):
```
ulfalizer: No leading underscores, since this is a local variable/function.

menuconfig.py
```diff
@@ -165,100 +165,276 @@
...
+
+        t = next(tok)
+        bg_color = __parse_color(tok, t)
+
+        t = next(tok)
+        if t != None:
+            raise StyleError("Garbage at end of string. Error near: {}".format(t))
+
+    attrs = 0
+    fg_color = -1
```
ulfalizer: Instead of having `fg_color` and `bg_color` default to -1 if not specified, it would be nice if they were inherited from e.g. the theme, like for the other attributes. That way all the attributes would work the same ways.

menuconfig.py
```diff
@@ -291,6 +292,10 @@ def style_from_string(sstr):
 
     return style
 
+def style_set(sstr):
```
ulfalizer: Think I'd prefer for there to just be a single mechanism, at least until we run into any limitations with it.

To set the style programmatically, you could always just set `os.environ["MENUCONFIG_STYLE"]` in the code.

menuconfig.py
```diff
@@ -267,123 +267,104 @@ def style_from_string(sstr):
...
     attrs = 0
     fg_color = -1
     bg_color = -1
 
-    __parse_main(__parse_tokenizer(cstr))
+    # Parse attributes
+    for t in cstr.split(','):
+        if t == '': continue
+
+        if t == 'bold':
```
ulfalizer: My preferred quoting style is `'bold'`.

Or to be honest, I don't really have a strong preference, but it's what the rest of the code already uses, and consistency is nice. :)

ulfalizer:
Could you merge all the commits into a single commit for future versions? It makes the history cleaner once stuff is merged, and makes things easier to review.

You can use `git push --force` to push with a rewritten history.

ulfalizer:
Thinking about it, your design for attributes (including the -1 defaults for colors) makes more sense.

If a later style does not specify `bold` for example, then you don't want the "bold" attribute to be inherited, because then there'd be no way to undo it. And if attributes like that won't be inherited, then colors probably shouldn't be inherited either. So `foo=...` should probably just replace the entire style for `foo`, like you did it.

Sorry for the noise!

To clarify the design I had in mind, this is what could happen internally if `MENUCONFIG_STYLE="separator=fg:white selection=bold"` is passed in:

1. `"default"` is added to the beginning of `MENUCONFIG_STYLE`, giving `"default separator=fg_white selection=bold"`

2. The style parsing code runs on the string and sees that `default` doesn't have a `=` in it, looks it up as a style, and parses its style string (which is just a normal `MENUCONFIG_STYLE` string). This fills in fields in a style dictionary, where the `default` style is guaranteed to specify values for all keys.

3. The style parsing code next sees the `separator=fg:white`, and replaces the style at the `separator` key in the dictionary.

4. The style parsing sees the `selection=bold` and similarly replaces the style at the `selection` key.

5. The style dictionary now holds the final style, which is initialized.

You could also have stuff like `MENUCONFIG_STYLE="separator=fg:white selection=bold aquatic"`, which would give all the attributes specified in the `aquatic` theme their values from that style.

pinkfluid:
Hi,

I did rebase-squash to a single commit. FYI, I think that github allows you to merge pull request in several different ways -- squash being one of them.

I did change the quotation style and prefixed '_' the global functions.

Regarding your last comment, I believe the code currently already works the way you described.

A style is basically a dictionary of keywords which override the keywords in the parent style (aka template). If no template is selected, the default style is used. If a keyword doesn't exists in the parent style, it is considered an error. This means that the default style also defines the list of valid keywords.

The only special keyword is "style", which selects the template.

The order of keywords in MENUCONFIG_STYLE doesn't really affect the result. Your example above (MENUCONFIG_STYLE="separator=fg:white selection=fg:white,bg:black,bold aquatic) is literally translated to the dictionary below:

```python
{
    "style": "aquatic",
    "separator": "fg:white"
    "selection": "fg:white,bg:black,bold"
}
```
What happens internally is this:

```
MENUCONFIG_STYLE                    _AQUATIC_STYLE = \ _                            _DEFAULT_STYLE = \
{                                   {                                               {
    "style": "aquatic",                 "style": "default",
                                        "path": "fg:cyan,bg:blue,bold",  ============>  "path": "fg:black,bg:white,bold",
    "separator": "fg:white",   ======>  "separator": "fg:white,bg:cyan,bold",  ======>  "separator": "fg:black,bg:yellow,bold",
                                                                                        "list": "fg:black,bg:white",
    "selection": "fg:white,bg:black,bold",  =========================================>  "selection": "fg:white,bg:blue,bold",
                                                                                        "inv_list": "fg:red,bg:white",
                                                                                        "inv_selection": "fg:red,bg:white",
                                        "help":  "fg:cyan,bg:blue,bold", ============>  "help": "fg:black,bg:white,bold",
                                        "frame": "fg:white,bg:cyan,bold",============>  "frame": "fg:black,bg:yellow,bold",
                                        "body": "fg:white,bg:blue", =================>  "body": "fg:white,bg:black",
                                        "edit": "fg:black,bg:white", ================>  "edit": "fg:white,bg:blue",
                                                                                        "jump_to_edit": "fg:white,bg:blue",
                                                                                        "text":  "fg:black,bg:white"
}                                   }                                               }

```
So basically that means that the default style is used as the template for aquatic, which is used as the template for the MENUCONFIG_STYLE.

This is the end result:


2018-09-01-214940_954x430_scrot

pinkfluid:
FYI I've been also investigating RGB color mode (for example, you could specify fg:#FFFFFF) and it is a bit trickier than initially thought. But I have experimental code that is able to use RGB mode if available, otherwise it can map a RGB value to the standard 256 colors (minus the first 16 colors as they are user defined and there's no way to query their RGB values).

There's also another color mode referred to as "true color" where you're able to use RGB colors without overwriting the current color scheme, but unfortunately ncurses doesn't support it.

menuconfig.py
```diff
@@ -165,100 +165,266 @@
...
+        "help":             "fg:cyan,bg:blue,bold",
+        "frame":            "fg:white,bg:cyan,bold",
+        "body":             "fg:white,bg:blue",
+        "edit":             "fg:black,bg:white",
+}
+
+_STYLES = {}
+_CUSTOM_STYLE = None
+
+class _StyleError(Exception):
```
ulfalizer: Should have `StyleError` as part of the public API too, so no leading `_`.

menuconfig.py
```diff
@@ -165,100 +165,266 @@
...
+    pass
+
+def _style_expand(sdef, base = None):
```
ulfalizer: No spaces around `=` for default values (quirky Python convention).

menuconfig.py
```diff
@@ -165,100 +165,266 @@
...
+{
+        # Top row in the main display, with the menu path
+        "path":             "fg:black,bg:white,bold",
+        # Separator lines between windows. Also used for the top line in the symbol
+        "separator":        "fg:black,bg:yellow,bold",
+        # List of items, e.g. the main display
+        "list":             "fg:black,bg:white",
+        # Style for the selected item
+        "selection":        "fg:white,bg:blue,bold",
+        # Like _LIST_(SEL_)STYLE, for invisible items. Used in show-all mode.
```
ulfalizer: Could say "Like list/selection" instead maybe.

menuconfig.py
```diff
@@ -165,100 +165,266 @@
...
+    attrs = 0
+    fg_color = -1
+    bg_color = -1
+
+    # Parse attributes
+    for t in cstr.split(","):
+        if t == "": continue
+
+        if t == "bold":
+            attrs |= curses.A_NORMAL if _IS_WINDOWS else curses.A_BOLD
```
ulfalizer: The comment explaining this ("A_BOLD tends to produce faint...") disappeared.

menuconfig.py
```diff
@@ -165,100 +165,266 @@
...
-    #   1. Text color
-    #   2. Background color
-    #   3. Attributes
-    #   4. Extra attributes if colors aren't available. The colors will be
-    #      ignored in this case, and the attributes from (3.) and (4.) will be
-    #      ORed together.
+    # Merge the two dictionaries
+    for x in sdef:
+        # Skip the "style" special keyword
+        if x == "style": continue
```
ulfalizer: Wouldn't be any need for the `style` keyword with the string-only design I think.

menuconfig.py
```diff
@@ -165,100 +165,266 @@
...
-    # Help text windows at the bottom of various fullscreen dialogs
-    HELP_STYLE               = PATH_STYLE
+        # Words without a "=" character imply a style selection
+        if len(skey) == 1:
+            # Implicity style selection
 
-    # Frame around dialog boxes
-    DIALOG_FRAME_STYLE       = SEPARATOR_STYLE
+            if not skey[0] in _STYLES:
+                raise _StyleError("Unknown built-in style: {}. Error near: {}".format(skey[0], sline))
```
ulfalizer: Could linebreak before the `.format()` to shorten the line a bit.

I try to stay within 80 columns.

menuconfig.py
```diff
@@ -165,100 +165,266 @@
...
+            style["style"] = skey[0]
 
-    # Edit box in pop-up dialogs
-    DIALOG_EDIT_STYLE        = (curses.COLOR_WHITE, curses.COLOR_BLUE,   curses.A_NORMAL, curses.A_STANDOUT)
+        # key=data value
+        elif len(skey) == 2:
+            # The default stlye is the root of all styles therefore it must define style elements.
+            # This servers a double purpose of having a list of all valid keywords for styling
+            if not skey[0] in _DEFAULT_STYLE:
+                raise _StyleError("Unkown keyword: {}. Error near: {}".format(skey[0], sline))
```
ulfalizer: Could linebreak before `.format()`.

menuconfig.py
pinkfluid: The string representation of a style (as defined, for example, in the environment) is converted to a dictionary type. The "style" keyword is a dictionary keyword -- it is used when there's a word without '=' in it in the string.

So for example:

MENUCONFIG_STYLE="aquatic"

is converted to:

{
      "style": "aquatic"
}

Probably this could be renamed to "template" to make it clearer what the intention is.

ulfalizer:
> The string representation could work, the only issue that I foresee is that the _parse_styles() function should somehow avoid infinite recursions (style A references style B which references styles C, which references style A).
> 
> I'll give it some thought, meanwhile I pushed a new commit that fixes the issues mentioned above.

Thanks for the fixes!

Could pass a list of all the parent styles to `_parse_style()` to detect infinite recursion. Probably not a huge deal to leave it out for an initial version though, since it's so obscure.

pinkfluid:
Ok, I've got rid of dictionaries. Styles are now represented as strings. A dictionary form is used only when merging the style template with the current style. The recursion bug is there, but python gracefully crashes if that happens.

I've also implemented a _parse_styles() function that parses the string style representation.

The style options were better documented.

menuconfig.py
```diff
@@ -165,100 +165,274 @@
...
 
-    # Like _LIST_(SEL_)STYLE, for invisible items. Used in show-all mode.
-    LIST_INVISIBLE_STYLE     = (curses.COLOR_RED,   curses.COLOR_WHITE,  curses.A_NORMAL, BOLD             )
-    LIST_INVISIBLE_SEL_STYLE = (curses.COLOR_RED,   curses.COLOR_BLUE,   BOLD,            curses.A_STANDOUT)
+    return (style, template_name)
 
-    # Help text windows at the bottom of various fullscreen dialogs
-    HELP_STYLE               = PATH_STYLE
+def _parse_style(style_str=True):
+    tsyle = {}
```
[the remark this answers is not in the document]
mitja-plume: Good catch #1

menuconfig.py
```diff
@@ -165,100 +165,274 @@
...
+        style = _parse_style("monochrome")
+    # Use the user-defined style from the environemnt
+    elif "MENUCONFIG_STYLE" in os.environ:
+        style = _parse_style(os.environ["MENUCONFIG_STYLE"])
+    # Use a custom style set from menuconfig(style=...)
+    elif _CUSTOM_STYLE:
+        style = _parse_style(_CUSTOM_STYLE)
+    # Use the default style
+    else:
+        style = _parse_style("defualt")
```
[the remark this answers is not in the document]
mitja-plume: Good catch #2.

menuconfig.py
```diff
@@ -165,100 +165,274 @@
...
+    pass
 
-    # Default styling. Themes can override these settings below.
+def _parse_style_single(style_str):
+    template_name = None
+    style = {}
 
-    # Top row in the main display, with the menu path
-    PATH_STYLE               = (curses.COLOR_BLACK, curses.COLOR_WHITE,  BOLD                              )
+    for sline in style_str.replace("\n", "").split(" "):
```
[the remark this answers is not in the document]
mitja-plume: Not sure if that's true:

```
Python 3.7.0 (default, Jul 15 2018, 10:44:58)
[GCC 8.1.1 20180531] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> """
... foo=bar
... 1=2
... a=b
... this is just a test
... the end!
... """.split("=")
['\nfoo', 'bar\n1', '2\na', 'b\nthis is just a test\nthe end!\n']
```

menuconfig.py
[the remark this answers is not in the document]
pinkfluid: Good catch #1.

menuconfig.py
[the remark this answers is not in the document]
pinkfluid: Good catch #2.

menuconfig.py
```diff
@@ -165,100 +165,274 @@
...
+
+        elif cdef.startswith("color"):
+            cnum = cdef[len("color"):]
+            if not cnum.isdigit():
+                raise StyleError(
+                        "Fixed color must be followed by a number. " \
+                        "Error near: {}".format(t))
+
+            cnum = int(cnum)
+            if cnum >= curses.COLORS or cnum < -1:
```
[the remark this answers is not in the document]
pinkfluid: Fixed.

menuconfig.py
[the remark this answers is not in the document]
pinkfluid: Not sure if this is true:

```
Python 3.7.0 (default, Jul 15 2018, 10:44:58)
[GCC 8.1.1 20180531] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> """
... foo=bar
... 1=2
... a=b
... this is just a test
... the end!
... """.split("=")
['\nfoo', 'bar\n1', '2\na', 'b\nthis is just a test\nthe end!\n']
```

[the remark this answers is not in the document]
ulfalizer: Think you forgot to do the initial `.split()` there, to get the individual assignments.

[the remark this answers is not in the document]
pinkfluid: Ah, you mean split() without arguments. Yeah, that one works. Good to know. Will incorporate that.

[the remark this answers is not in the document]
ulfalizer: No need to `strip()` things separately when using a plain `split()` btw:

```python
>>> """
... foo=bar
...   baz=qaz
... hmm=zmh    weird=formatting
... """.split()
['foo=bar', 'baz=qaz', 'hmm=zmh', 'weird=formatting']
```

menuconfig.py
```diff
@@ -165,100 +165,274 @@
...
-    HELP_STYLE               = PATH_STYLE
+def _parse_style(style_str=True):
+    tsyle = {}
+    tname = None
 
-    # Frame around dialog boxes
-    DIALOG_FRAME_STYLE       = SEPARATOR_STYLE
+    # Parse the current style, tname will hold the style template
+    # name
+    (style, tname) = _parse_style_single(style_str)
```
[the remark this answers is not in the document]
pinkfluid: There are still two functions, mainly because we have to deal with the "default" theme being the starting point of all themes.

I did change the algorithm according to your suggestion, with a slight modification. I believe that the following examples should produce the same style, but correct me if I'm wrong:

MENUCONFIG_STYLE="aquatic selection=fg:red,bg:black"
vs
MENUCONFIG_STYLE="selection=fg:red,bg:black aquatic"

The second statement should not overwrite the "selection" item from aquatic.

Anyways, it's possible to achieve quite funny results right now:

MENUCONFIG_STYLE="monochrome aquatic"

2018-09-04-105235_1904x743_scrot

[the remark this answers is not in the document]
ulfalizer: > There are still two functions, mainly because we have to deal with the "default" theme being the starting point of all themes.

Could handle that with a hardcoded `_parse_style("default")` before doing anything else perhaps.

> I did change the algorithm according to your suggestion, with a slight modification. I believe that the following examples should produce the same style, but correct me if I'm wrong:
> 
> MENUCONFIG_STYLE="aquatic selection=fg:red,bg:black"
> vs
> MENUCONFIG_STYLE="selection=fg:red,bg:black aquatic"

IMO, it's fine if they produce different results. Always having stuff further to the right override stuff on the left makes the behavior simple to understand and flexible.

I'd want to document that it works like that later. I could finish up the doc part though.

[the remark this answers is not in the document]
pinkfluid: I'll change it to what you suggested -- actually that makes the code even simpler.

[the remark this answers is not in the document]
pinkfluid: Fixed.

menuconfig.py
```diff
@@ -165,100 +165,255 @@
...
+                raise StyleError("Unknown built-in style: {}. Error near: {}".format(
+                    skey[0], sline))
+
+            # Merge in the template
+            tstyle = _parse_style_expand(_STYLES[skey[0]])
+            style.update(tstyle)
+
+        # Key=data value, store it
+        elif len(skey) == 2:
+            style[skey[0]] = skey[1]
```
ulfalizer: Could move the `_style_to_curses()` call to here, to get rid of the calls in `_init_styles()`.

Since `_style_to_curses()` should already detect an error if `skey[1]` has an extra `=` in it, `_parse_style_expand()` could then be simplified too:

```
for sline in style_str.split():
    if "=" in sline:
        key, val = sline.split("=")
        style[key] = _style_to_curses(val)

    else:
        ...
        style.update(_parse_style_expand(_STYLES[sline]))

File: menuconfig.py [hunk 25]
```diff
@@ -165,100 +165,255 @@
...
+    global _LIST_STYLE
+    global _LIST_SEL_STYLE
+    global _LIST_INVISIBLE_STYLE
+    global _LIST_INVISIBLE_SEL_STYLE
+    global _HELP_STYLE
+    global _DIALOG_FRAME_STYLE
+    global _DIALOG_BODY_STYLE
+    global _DIALOG_EDIT_STYLE
+    global _JUMP_TO_EDIT_STYLE
+    global _INFO_TEXT_STYLE
```
ulfalizer: Guess these "bounce" variables are a bit silly now. Could just have a global `_style` dictionary instead, and reference that directly in the code.

I could fix that myselt though. Bit sadistic to force you to change a ton of unrelated code.

menuconfig.py
```diff
@@ -331,12 +486,14 @@ def menuconfig(kconf):
     global _config_filename
     global _show_all
     global _conf_changed
+    global _CUUSTOM_STYLE
```
ulfalizer: Typo: "_CUUSTOM_STYLE"

menuconfig.py
```diff
@@ -320,7 +475,7 @@ def _expr_str(expr):
 def _main():
     menuconfig(standard_kconfig())
 
-def menuconfig(kconf):
+def menuconfig(kconf, style=None):
```
ulfalizer: Do you need the extra `style` argument here?

I think I'd prefer if `MENUCONFIG_STYLE` was the only mechanism for setting the style (it could still be set programmatically via `os.environ`). If it turns out to be limiting in some way later, we could add the argument.

Some earlier experience has taught me that once you add something, you're stuck with it forever, even if it wasn't really needed. Best to add just the minimal interface at first. :)

menuconfig.py
pinkfluid: Fixed.

menuconfig.py
pinkfluid: Shouldn't be too hard, I'll give it a shot.

pinkfluid: Replaced all _*_STYLE variables with a single global _style dictionary.

ulfalizer: Thanks for fixing it for me!

Will take a look at the rest tomorrow.

ulfalizer: One thing you could do in the meantime if you want:

The final commit message should just describe the final change. No need to mention the fixes from earlier versions.

ulfalizer: Might look tomorrow instead. Pretty busy day today. :/

menuconfig.py
```diff
@@ -165,101 +165,221 @@
...
+            style.update(tstyle)
 
-    # Default styling. Themes can override these settings below.
+    return style
 
-    # Top row in the main display, with the menu path
-    PATH_STYLE               = (curses.COLOR_BLACK, curses.COLOR_WHITE,  BOLD                              )
+def _parse_style(style_str):
+    # Start off by using the default template
+    style = _parse_style_expand(_STYLES["default"])
```
ulfalizer: Just `_parse_style_expand("default")` should work here.

pinkfluid:
Ok, I believe I have addressed the latest batch of fixes. Thank you for reviewing this.

I've updated the docstring to describe the styling feature a bit better -- along these lines I did update the commit message to reflect the global status of the change.

There are two changes that you did not request:
1. I renamed the _style() function to _style_attr() as it was conflicting with the global _style dictionary
2. I've changed the "body" foreground color of the aquatic style from "white" to "brightwhite" as (imho) looks a bit better

menuconfig.py
```diff
@@ -48,13 +48,61 @@
...
 by Mitja Horvat (pinkfluid).
 
-See the _init_styles() function if you want to add additional themes. I'm happy
-to take them in upstream.
-
+This is the current list of built-in styles:
+    - default       classic Kconfiglib theme with a yellow accent
+    - monochrome    colorless theme (uses only bold and standout) attributes,
+                    this style is used if the terminal doesn't support colors
+    - aquatic       blue titnted style loosely resembling the lxdialog theme
```
ulfalizer: *tinted

menuconfig.py
```diff
@@ -48,13 +48,61 @@
...
+
+                    Note: On some terminals a bright version of the color implies
+                    bold.
+    - bold          Use bold text
+    - underline     Use underline text
+    - standout      Standout text attribute (reverse color)
+
+A keyword without the '=' is assumed to be a style template. The template name
+is looked up in the built-in styles list and the style definition is expanded
+in-place. With this, built-in stlyes can be used as basis for new styles.
```
ulfalizer: *styles

menuconfig.py
```diff
@@ -48,13 +48,61 @@
...
+
+For example, take the aquatic theme and give it a red selection bar:
+
+MENUCONFIG_STYLE="aquatic selection=fg:white,bg:red"
+
+If there's an error in the style definition, a SyleError exception will be
```
ulfalizer: *StyleError

menuconfig.py
```diff
@@ -48,13 +48,61 @@
...
+thrown briefly describing the error.
```
ulfalizer: Maybe something like the following could be inserted here:

```diff
 
+The 'default' theme is always implicitly parsed first (or the 'monochrome'
+theme if the terminal lacks colors), so the following two settings have the
+same effect:
+
+    MENUCONFIG_STYLE="selection=fg:white,bg:red"
+    MENUCONFIG_STYLE="default selection=fg:white,bg:red"
+
+
 Other features
```

menuconfig.py
```diff
@@ -165,101 +213,172 @@
...
-    # Frame around dialog boxes
-    DIALOG_FRAME_STYLE       = SEPARATOR_STYLE
+_STYLES = {
+    "default": """
+    path=fg:black,bg:white,bold
+    separator=fg:black,bg:yellow,bold
+    list=fg:black,bg:white
+    selection=fg:white,bg:blue,bold
+    inv-list=fg:red,bg:white
+    inv-selection=fg:red,bg:white
```
ulfalizer: Should be `fg:red,bg:blue`, or the cursor disappears.

menuconfig.py
```diff
@@ -270,17 +389,10 @@ def _init_styles():
...
-           color_attribs={}):
+def _style_attr(fg_color, bg_color, attribs, color_attribs={}):
     # Returns an attribute with the specified foreground and background color
     # and the attributes in 'attribs'. Reuses color pairs already created if
     # possible, and creates a new color pair otherwise.
-    #
-    # Returns 'attribs | no_color_extra_attribs' if colors aren't supported.
-
-    if not curses.has_colors():
-        return attribs | no_color_extra_attribs
```
ulfalizer: This is still needed, because `curses.init_pair()` throws an exception on terminals that don't support color.

A good test is to turn with `TERM=vt100`.

ulfalizer: (Won't run into many terminals that don't support color these days, but sometimes wonky stuff can happen with people's TERM settings, and it's nice to not crash then.)

pinkfluid: That was actually a derp on my side ... doing too many things at once isn't always a good idea :)

Fixed.

menuconfig.py
pinkfluid: Fixed.

menuconfig.py
pinkfluid: Yeah, makes sense. Added.

pinkfluid:
Fixed the typos and monochrome terminals. Pushed new version.

ulfalizer:
Finally merged. Thanks a lot!

Should be pushing out a new version with it soon.

ulfalizer:
I made a small tweak to it to make non-predefined colors a bit easier to use: https://github.com/ulfalizer/Kconfiglib/commit/0bb65bc5628c7493897bf2f8cc786bcbbbcdefe5

Any objections?

pinkfluid:
Finally, thank you very much :)

One more thing... I've been investigating the possibility of adding RGB color support (for example, by using HTML notation -- fg:#ffffff,bg:#000000) and it seems to be more complicated than I initially thought.

There are basically 4 color modes on modern terminals:

- basic 16 color mode
- 256 color mode
- RGB mode
- TrueColor mode

16 color mode is self explanatory, while TrueColor mode is not support by ncurses, so lets ignore it for now.

256 color mode is basically an extension to the 16 color mode where the first 16 colors can be customized, while the colors above 15 are pre-defined.

Colors from 16-231 are a 6x6x6 RGB cube (values from 0-5 don't exactly translate to a RGB value from 0-255, but lets leave this for now). The colors from 232-255 is grey scale palette.

RGB mode is an extension to the 256 color mode. You can basically take any color from 0-255 and define its RGB value. So you can use 16 million colors, but only 256 at a time. This is pretty much how old VGA graphics card used to work.

So there are two options how to do this:

1. Support only 256 color mode - take a RGB value and map it to the closest color in the 256 color space. This is probably the simplest mode. For example, tmux uses this option.
2. Support RGB mode - this option needs option 1) as fallback for terminals that do not support RGB mode. What makes it even more complicated is that you cannot mix the 256 color notation (colorXXX) with the RGB color notation (#RRGGBB) as the RGB notation will be overwriting colors in the 256 color space. To have both, you basically need to emulate 256-color mode with RGB mode (colorXXX -> convert to the RGB value -> allocate RGB color).

Now, option 2. sounds complicated but I already have python code that is able to do that (256 emulation + RGB allocation). I compared it to tmux and the results are comparable if not slightly better (mainly because tmux converts a color to grey scale by averaging the R,G,B values, which is not exactly correct).

The real question is -- is this something you might want in kconfiglib? My rough estimate is that option 1) is about 50-80 LOC, while option 2 is 150-200.

Frankly, this feature may not be worth the extra LOCs needed to implemented it, but if you feel it is something that might be useful, I can share the code or maybe come up with another pull request. Option 1 should be simple enough though.

Please share your thoughts.

pinkfluid:
Regarding the colorNNN -> N change, no objections, looks good.

ulfalizer:
Added an error-related change too: https://github.com/ulfalizer/Kconfiglib/commit/4e25d1c713b0c37feb6a8781a92b62a52ee5eb04

Thought it might be mean to not point out missing styles/style templates at least, even if it's not an error.

ulfalizer:
Yeah, terminals are surprisingly messy for a grid of text...

I dug around a bit. Are you sure there are terminals that support 256 colors but only allow you to change the first 16? The 256-color stuff is an `xterm` extension originally and not implemented on any real hardware terminals, so it seems a bit arbitrary to limit changing colors to the first 16 colors if you allow changing colors at all.

On the `curses` side, there's just `has_colors()` for checking if you can use colors, and `can_change_color()` for checking if you can modify the palette entries (and `COLORS` for checking how many colors are in the palette).

Could take a look at whatever approach you prefer if you put together a proof-of-concept. Weekend mode initiated now though, so might be off for a while. :)

pinkfluid:
No, I didn't meant you can change only the first 16 colors, but the other way around :)

In RGB mode you can change all colors from 0..curses.COLORS. The problem with the first 16 colors is that typically users customize those. And  there's no way to query the current RGB value of a colors.

The colors from 16-255 can be calculated using a formula (or it might be that curses.color_content() can be used for that, so we can just build a table at startup), so even if we overwrite those RGB values we can always recalculate them.

The algorithm would roughly look like this:
```
# Skip first 16 colors
num_rgb_colors = 15

def get_colorrgb(r, g, b):
        global num_rgb_colors
        num_rgb_colors += 1
        curses.init_color(num_rgb_colors, r, g, b)
        return num_rgb_colors

# Get a color from the standard 256 color palette
def get_color256(int num):
    if num < 16: return num

   (r, g, b) = caluclate_rgb_from_standard_color_using_formula(num)
   return get_colorrgb(r, g, b)

```
I'll send a pull request in the following days; but as I said, it could be too much code for just a little feature :)

PR #55, opened 2018-09-10
menuconfig: Add RGB color support to styles.

pinkfluid:
Styles can now take a HTML notation as a color value:
- on terminals that support RGB color mode these are rendered
  perfeclty
- on 256 color terminals the closes matching color is used
- on other terminals, the current color palette is scanned
  and the closes matching color is used. This has few quirks

Example: Use "pure" red as selection foreground and "pure" blue as
background:

MENUCONFIG_STYLE="fg:#ff0000,bg:#0000ff"

Afterall, didn't turn out too bad. About 170 lines with comments.

Please review and decide if it is worthwhile to include. May still bit a bit overkill for this feature.

### DIFF
--- a/menuconfig.py
+++ b/menuconfig.py
@@ -87,7 +87,11 @@ The color definition is a comma separated list of attributes:
                     you can also directly put in a color number, e.g. fg:123
                     (hexadecimal and octal constants are accepted as well).
                     Colors outside the range -1..curses.COLORS-1 (which is
-                    terminal-dependent) are ignored (with a warning).
+                    terminal-dependent) are ignored (with a warning). The COLOR
+                    can be also specified using a RGB value in the HTML
+                    notation, for example #RRGGBB. If the terminal supports
+                    color changing, the color is render accurately. Otherwise
+                    the visually nearest color is used.
 
                     If the background or foreground color of an element is not
                     specified, it defaults to -1, representing the default
@@ -98,6 +102,11 @@ The color definition is a comma separated list of attributes:
     - bold          Use bold text
     - underline     Use underline text
     - standout      Standout text attribute (reverse color)
+
+More often than not, some UI elements share the same color definition. In such
+cases the right value may specify an UI element from which the color definition
+will be copied. For example, "separator=help" will apply the current color
+definition for "help" to "separator".
 
 A keyword without the '=' is assumed to be a style template. The template name
 is looked up in the built-in styles list and the style definition is expanded
@@ -299,6 +308,166 @@ _STYLE_STD_COLORS = {
     "brightpurple": curses.COLOR_MAGENTA + 8,
 }
 
+def _rgb_to_6cube(r, g, b):
+    # Take an 888 RGB color value and return a 3 tuple representing the index
+    # in the xterm 6x6x6 color cube
+
+    # Xterm uses a RGB color palette, where the values of each component
+    # can be between 0-5 (therefore forming a 6x6x6 cube). The catch is
+    # that the mapping between RGB and the 6x6x6 cube is non-linear. The
+    # 6x6x6 cube index of 0 is mapped to a RGB value of 0. 1-5 are mapped
+    # to RGB values of 95 with increments of 40.
+
+    # Sources:
+    # https://commons.wikimedia.org/wiki/File:Xterm_256color_chart.svg
+    # https://github.com/tmux/tmux/blob/master/colour.c
+
+    # The formula for converting a single RGB value to the 6cube index
+    # 48 is the middle ground between 0 and 95
+    rgb_to_c6 = lambda x: 0 if x < 48 else max(1, int(round((x - 55) / 40)))
+
+    return rgb_to_c6(r), rgb_to_c6(g), rgb_to_c6(b)
+
+def _rgb_from_6cube(r6, g6, b6):
+    # Take a 666 xterm color cube index and convert it to a
+    # 3 tuple representing the RGB color
+
+    # The formula from converting the xterm 6cube index to a RGB value
+    rgb_from_c6 = lambda x: 0 if x == 0 else 55 + x*40
+
+    return rgb_from_c6(r6), rgb_from_c6(g6), rgb_from_c6(b6)
+
+def _rgb_to_gray_idx(r, g, b):
+    # Convert an 888 RGB color to the index of an xterm 256-color grayscale color
+    # with approx. the same perceived brightness. This "grayscale candidate" can
+    # be compared against the best "color candidate" to find a good xterm color
+    # to represent the color.
+
+    # Calculate the luminance (gray intensity) of a color from its R, G, B components
+    # Source: https://stackoverflow.com/questions/596216/formula-to-determine-brightness-of-rgb-color
+    luma = 0.299 * r + 0.587 * g + 0.114 * b
+
+    # Closests index in the grayscale palette, which starts at RGB 0x080808, with
+    # stepping 0x0A0A0A
+    idx = int(round((luma - 0x08) / 0x0A))
+
+    # Clamp the index to 0-23, corresponding to 232-255
+    return max(0, min(idx, 23))
+
+def _rgb_from_gray_idx(c):
+    # Convert a grayscale index to its closet single RGB component
+
+    g = c * 10 + 8
+    return (g, g, g)
+
+# Obscure Python: rgb2index is initialized from a reference to a global {}
+# Modification to this dictionary are retained between calls _rgb_alloc(), thus
+# making rgb2index a static variable
+def _rgb_alloc(rgb, rgb2index={}):
+    # Initialize a new entry in the xterm palette to the given RGB color, returning its
+    # index. If the color has already been initialized, the index of the existing entry is
+    # returned.
+
+    # Ncurses doesn't allow you to define new colors -- you are allowed only
+    # to overwrite existing ones.
+
+    # The colors from 0-15 are user-defined and there's no way to query the
+    # RGB values so we better leave them untouched.
+
+    # The RGB values of colors from 16-255 can be easily calculated. However,
+    # colors from 232-255 use a different formula (grayscale) so stick with
+    # colors from 16-231. 200+ colors should be plentiful.
+
+    if rgb in rgb2index:
+        return rgb2index[rgb]
+
+    # The first 16 colors are user defined and we should not change their
+    # values
+    cn = len(rgb2index) + 16
+    if cn >= curses.COLORS:
+        _warn("Unable to allocate new RGB color.", rgb)
+        return 0
+
+    # Map each RGB component from the range 0-255 to the range 0-1000, which
+    # is what curses uses
+    curses.init_color(cn, *(int(round(x * 1000 / 255)) for x in rgb))
+    rgb2index[rgb] = cn
+
+    return cn
+
+def _color_get(num):
+    # Returns the index of a color that looks like color 'num' in the xterm
+    # 256-color palette.
+
+    # If ncurses supports color changes, We can't return 'num' directly when
+    # redefining colors, since we might have overwritten the palette entry at
+    # 'num'. Instead, we allocate a new color for it, emulating the 256-color
+    # palette.
+
+    # Simplest case -- terminal doesn't support changing the definition of
+    # colors. _color_get_rgb() won't be changing the current palettte
+    # so we can return the color as-is
+    if not curses.can_change_color():
+        return num
+
+    # Standard colors, _rgb_alloc() doesn't touch these so
+    # we can return them as-is
+    if num < 16 or num >= 232:
+        return num
+
+    # We're in RGB mode. _rgb_alloc() will redefine the colors
+    # from 16-231. Since the RGB value of the "standard" 256 color
+    # palette can be easily calculated, we can emulate them in RGB
+    # mode rather easily
+    num -= 16
+    return _rgb_alloc(_rgb_from_6cube((num // 36) % 6, (num // 6) % 6, num % 6))
+
+def _color_get_rgb(rgb):
+    # Lambda for calculating the Euclidean distance between two RGB colors
+    dist = lambda r1, r2: sum((x - y)**2 for x, y in zip(r1, r2))
+
+    # Best case -- terminal supports the changing of colors
+    if curses.COLORS >= 256 and curses.can_change_color():
+        return _rgb_alloc(rgb)
+    # Second best case -- terminal supports 256 colors
+    # Find the closes matching color in the standard 6x6x6 color palette and
+    # the greyscale palette (232-255), compare the two and select the closest
+    # matching color
+    elif curses.COLORS >= 256:
+        # Calculate the indexes of the closest RGB color in the color palette
+        c6 = _rgb_to_6cube(*rgb)
+        # Calculate the RGB value of the closest color
+        crgb = _rgb_from_6cube(*c6)
+
+        # Calculate the index value of the closest gray palette color
+        gr = _rgb_to_gray_idx(*rgb)
+        # Calculate back RGB value
+        cgr = _rgb_from_gray_idx(gr)
+
+        if dist(rgb, crgb) < dist(rgb, cgr):
+            # Use the 6x6x6 color palette, calculate the color number
+            # from the 6cube index triplet
+            return 16 + (c6[0] * 36) + (c6[1] * 6) + c6[2]
+        else:
+            # Use the gray palette
+            return 232 + gr
+
+    # No support for color changes or 256 color mode, this is probably the best
+    # we can do, or is it? Submit patches :)
+    color = 0
+    dmin = 255**2 + 255**2 + 255**2
+    for x in range(0, curses.COLORS):
+        crgb = curses.color_content(x)
+        # ncurses returns colors with a range from 0..1000, scale that down
+        # to 0..255
+        crgb = [int(x * 255 / 1000) for x in crgb]
+        d = dist(rgb, crgb)
+        if d < dmin:
+            dmin = d
+            color = x
+
+    return color
+
 # Dictionary mapping element types to the curses attributes used to display
 # them
 _style = {}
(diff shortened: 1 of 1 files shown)


menuconfig.py
```diff
@@ -299,6 +299,141 @@
...
+    g = c * 10 + 8
+    return (g, g, g)
+
+def rgb_alloc(rgb, RGB_COLORS={}):
+    """
+    Ncurses doesn't allow you to define new colors -- you are allowed only
+    to overwrite existing ones.
+
+    The colors from 0-15 are user-defined and there's no way to query the
+    RGB values so we better leave them untouched.
```
ulfalizer: Do you have a reference for being unable to query the RGB values for colors in the range 0-15?

Even 0-15 is an extension by the way (from [aixterm](https://invisible-island.net/ncurses/ncurses.faq.html) it seems). ANSI only defined eight colors. :)

menuconfig.py
```diff
@@ -299,6 +299,141 @@
...
+    colors from 16-231. 200+ colors should be plentiful.
+    """
+    if rgb in RGB_COLORS:
+        return RGB_COLORS[rgb]
+
+    # The first 16 colors are user defined and we should not change their
+    # values
+    cn = len(RGB_COLORS) + 16
+    if cn >= curses.COLORS:
+        raise StyleError("Unable to allocate new RGB color.")
```
ulfalizer: `StyleError` no longer exists. Use `_warn()` instead.

A messed-up display is probably nicer than a hard error.

menuconfig.py
```diff
@@ -299,6 +299,141 @@
...
+        return num
+
+    # We're in RGB mode. rgb_alloc() will redefine the colors
+    # from 16-231. Since the RGB value of the "standard" 256 color
+    # palette can be easily calculated, we can emulate them in RGB
+    # mode rather easily
+    num -= 16
+    rgb = rgb_from_6cube((num / 36) % 6, (num / 6) % 6, num % 6)
+
+    return rgb_alloc(rgb)
```
ulfalizer: Could have `return rgb_alloc(rgb_from_6cube(...))`.

menuconfig.py
```diff
@@ -299,6 +299,141 @@
...
+
+def color_get_rgb(rgb):
+    # Lambda for calculating the distance between two RGB colors
+    dist = lambda r1, r2: sum([(x - y)**2 for x, y in zip(r1, r2)])
```
ulfalizer: Could have `sum((x - y)**2 for x, y in zip(r1, r2))`, using a generator instead of a list comprehension.

Maybe the comment could say `...for calculating the (squared) distance...`, just to be clear.

menuconfig.py
```diff
@@ -299,6 +299,141 @@
...
+        # Calculate the RGB value of the closest color
+        crgb = rgb_from_6cube(c6)
+
+        # Calculate the index value of the closest gray palette color
+        gr = rgb_to_gridx(rgb)
+        # Calculate back RGB value
+        cgr = rgb_from_gridx(gr)
+
+        if dist(rgb, crgb) < dist(rgb, cgr):
+            # Use the 6x6x6 color palette, calcualte the color number
```
ulfalizer: *calculate

menuconfig.py
```diff
@@ -299,6 +299,141 @@
...
+    # Formula from converting a RGB value to a curses color
+    # It's really weird that they chose 1000 as a max value
+    ctocr = lambda x: int(round(x * 1000 / 255))
+    curses.init_color(cn, *[ctocr(x) for x in rgb])
+    RGB_COLORS[rgb] = cn
+
+    return cn
+
+def color_get(num):
+    # Simplest case -- terminal doesn't support RGB colors
```
ulfalizer: Instead of "RGB colors", could it say "doesn't support changing the definitions of colors" (and same thing elsewhere too)? Makes it a bit more concrete.

menuconfig.py
```diff
@@ -299,6 +299,141 @@
...
+    # so we can return the color as-is
+    if not curses.can_change_color():
+        return num
+
+    # Standard colors, rgb_alloc() doesn't touch these so
+    # we can return them as-is
+    if num < 16 or num >= 232:
+        return num
+
+    # We're in RGB mode. rgb_alloc() will redefine the colors
```
ulfalizer: Not sure if RGB mode is a standard phrase. Could say something like "we can change the definitions of colors".

menuconfig.py
```diff
@@ -299,6 +299,141 @@
...
+    # values
+    cn = len(RGB_COLORS) + 16
+    if cn >= curses.COLORS:
+        raise StyleError("Unable to allocate new RGB color.")
+
+    # Formula from converting a RGB value to a curses color
+    # It's really weird that they chose 1000 as a max value
+    ctocr = lambda x: int(round(x * 1000 / 255))
+    curses.init_color(cn, *[ctocr(x) for x in rgb])
+    RGB_COLORS[rgb] = cn
```
ulfalizer: Maybe `RGB_COLORS` could be called `rgb2index` or the like.

menuconfig.py
```diff
@@ -299,6 +299,141 @@
     "brightpurple": curses.COLOR_MAGENTA + 8,
 }
 
+def rgb_to_6cube(rgb):
+    """
+    Convert each component of a RGB color to its approximate index
+    in the xterm 6cube color palette
+    """
+    # The formula for converting a single RGB value to the 6cube index
+    rto6 = lambda x: 0 if x < 48 else max(1, int(round((x - 55) / 40)))
```
ulfalizer: Would be nice to either explain the formula or include a link to some reference here.

Can just say "Converts an RGB value to a 6cube index. (Explanation ...)" too.

menuconfig.py
```diff
@@ -299,6 +299,141 @@
     "brightpurple": curses.COLOR_MAGENTA + 8,
 }
 
+def rgb_to_6cube(rgb):
```
ulfalizer: Not sure if good idea, but maybe these functions could take three parameters instead.

menuconfig.py
```diff
@@ -299,6 +299,141 @@
...
+    """
+    Convert each component of a RGB color to its approximate index
+    in the xterm 6cube color palette
+    """
+    # The formula for converting a single RGB value to the 6cube index
+    rto6 = lambda x: 0 if x < 48 else max(1, int(round((x - 55) / 40)))
+
+    return [rto6(x) for x in rgb]
```
ulfalizer: With three parameters, this would become

    return (rto6(r), rto6(g), rto6(b))

Seems pretty nice and explicit, and you'd consistently deal with tuples.

ulfalizer:
Ahh... looking at https://upload.wikimedia.org/wikipedia/commons/1/15/Xterm_256color_chart.svg, I see that it's offset at the beginning (jumps directly to 95 instead of 40), so that explains the 55.

Could you briefly explain tricky stuff like that in the comments (or link a reference)?

menuconfig.py
```diff
@@ -299,6 +299,142 @@
...
+    # The formula from converting the xterm 6cube index to a RGB value
+    rfrom6 = lambda x: 0 if x == 0 else 55 + x*40
+    return [rfrom6(x) for x in c6]
+
+def rgb_to_gridx(rgb):
+    """
+    Convert a RGB color to its closest xterm grayscale color
+    (indexes from 0-23 represents xterm colros 232-255)
+    """
+    gray = int(round((rgb[0] + rgb[1] + rgb[2]) / 3))
```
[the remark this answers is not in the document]
ulfalizer: Never mind... think I follow the logic now.

Is this a somewhat correct description?

    Convert an RGB color to an xterm 256-color grayscale color that would
    represent it well if we assume that it is already close to grayscale. This
    "grayscale candidate" can be compared against the best "color candidate" to
    find a good xterm color to represent the color.

"Closest" confused me, because I think you'd need to do something fancier to find the grayscale color that you perceive as similarly bright.

menuconfig.py
ulfalizer: Hmm... yeah, doesn't seem to be possible with gnome-terminal at least (get a fixed value from `color_content()` even after changing a color in the UI), and it feels a bit shaky to rely on anyway.

menuconfig.py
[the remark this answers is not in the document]
ulfalizer: Just realized that this also gives the gray with the shortest Euclidean distance. I should do math stuff more often. :)

Is there any point to truncating the value already in the beginning? Maybe it could be something like this:

```python
# Closest gray (in an Euclidean sense, projecting onto the line r = g = b)
gray = (rgb[0] + rgb[1] + rgb[2])/3

# Closests index in the grayscale palette, which starts at RGB 0x080808, with
# stepping 0x0A0A0A
gray_index = int(round((gray - 8)/10))

# Clamp the index to [0, 23] (corresponding to 232-255)
return max(0, min(gray_index, 23))
```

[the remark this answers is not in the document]
pinkfluid: There's a fancier formula for finding the intensity, and it's what I originally used in my prototype:
round(rgb[0] * 0.299 + rgb[1] * 0.587 + rgb[2] * 0.114)

We could still use that, it will probably give better accuracy for matching the shades of gray. I did do a visual comparison with tmux, and seemed a bit better at picking proper gray shades.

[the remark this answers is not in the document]
pinkfluid: Source: https://stackoverflow.com/questions/596216/formula-to-determine-brightness-of-rgb-color

menuconfig.py
```diff
@@ -299,6 +299,142 @@
...
+    Convert a 6cube index triplet to the corresponding RGB value
+    """
+    # The formula from converting the xterm 6cube index to a RGB value
+    rfrom6 = lambda x: 0 if x == 0 else 55 + x*40
+    return [rfrom6(x) for x in c6]
+
+def rgb_to_gridx(rgb):
+    """
+    Convert a RGB color to its closest xterm grayscale color
+    (indexes from 0-23 represents xterm colros 232-255)
```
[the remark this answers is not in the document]
pinkfluid: No, gridx simply stands for gray index. I'll update the function name.

menuconfig.py
pinkfluid: Changed to use the term "Euclidean" distance.

pinkfluid:
I think I've addressed most of the issues above, pushed a new squashed commit.

menuconfig.py
```diff
@@ -299,6 +299,159 @@
...
+    """
+    g = c * 10 + 8
+    return (g, g, g)
+
+# Obscure Python: rgb2index is initialized from a reference to a global {}
+# Modification to this dictionary are retained between calls rgb_alloc(), thus
+# making rgb2index a static variable
+def rgb_alloc(rgb, rgb2index={}):
+    """
+    Ncurses doesn't allow you to define new colors -- you are allowed only
```
ulfalizer: Might be nice to start the docstring with a high-level description of what the function does:

    Initialize a new entry in the xterm palette to the given RGB color, returning its
    index. If the color has already been initialized, the index of the existing entry is
    returned.

    Ncurses doesn't allow...

menuconfig.py
```diff
@@ -299,6 +299,159 @@
...
+    colors from 16-231. 200+ colors should be plentiful.
+    """
+    if rgb in rgb2index:
+        return rgb2index[rgb]
+
+    # The first 16 colors are user defined and we should not change their
+    # values
+    cn = len(rgb2index) + 16
+    if cn >= curses.COLORS:
+        _warn("Unable to allocate new RGB color.")
```
ulfalizer: Could make it `_warn("Unable to allocate new RGB color", rgb)`, just to help a bit with debugging.

menuconfig.py
```diff
@@ -299,6 +299,159 @@
...
+        return 0
+
+    # Map each RGB component from the range 0-255 to the range 0-1000, which
+    # is what curses uses
+    curses.init_color(cn, *(int(round(x * 1000 / 255)) for x in rgb))
+    rgb2index[rgb] = cn
+
+    return cn
+
+def color_get(num):
```
ulfalizer: Docstring (or "doc comment") could be this:

    # Returns the index of a color that looks like color 'num' in the xterm 256-color palette.
    #
    # We can't return 'num' directly when redefining colors, since we might have overwritten
    # the palette entry at 'num'. Instead, we allocate a new color for it, emulating the
    # 256-color palette.

menuconfig.py
```diff
@@ -299,6 +299,159 @@
...
+    # We're in RGB mode. rgb_alloc() will redefine the colors
+    # from 16-231. Since the RGB value of the "standard" 256 color
+    # palette can be easily calculated, we can emulate them in RGB
+    # mode rather easily
+    num -= 16
+    return rgb_alloc(rgb_from_6cube((num / 36) % 6, (num / 6) % 6, num % 6))
+
+def color_get_rgb(rgb):
+    # Lambda for calculating the Euclidean distance between two RGB colors
+    dist = lambda r1, r2: sum(((x - y)**2 for x, y in zip(r1, r2)))
```
ulfalizer: Generator expressions don't need parenthesis inside parameter lists, so could make it

    dist = lambda r1, r2: sum((x - y)**2 for x, y in zip(r1, r2))

ulfalizer:
One more thing:

These functions are internal, so they should be prefixed with `_`.

pinkfluid:
Sorry for the late response, but it was pretty hectic today. Thanks for the documentation/comment fixes. I believe I've integrated all of your suggestion in the new squashed commit.

menuconfig.py
```diff
@@ -299,6 +299,173 @@
...
+
+def _rgb_from_gray_idx(c):
+    """
+    Convert a grayscale index to its closet single RGB component
+    """
+    g = c * 10 + 8
+    return (g, g, g)
+
+# Obscure Python: rgb2index is initialized from a reference to a global {}
+# Modification to this dictionary are retained between calls rgb_alloc(), thus
```
ulfalizer: *`_rgb_alloc()`

menuconfig.py
```diff
@@ -299,6 +299,173 @@
...
+    palette.
+    """
+
+    # Simplest case -- terminal doesn't support changing the definition of
+    # colors. _color_get_rgb() won't be changing the current palettte
+    # so we can return the color as-is
+    if not curses.can_change_color():
+        return num
+
+    # Standard colors, rgb_alloc() doesn't touch these so
```
ulfalizer: *`_rgb_alloc()`

menuconfig.py
```diff
@@ -299,6 +299,173 @@
...
+    # we can return them as-is
+    if num < 16 or num >= 232:
+        return num
+
+    # We're in RGB mode. rgb_alloc() will redefine the colors
```
ulfalizer: *`_rgb_alloc()`

menuconfig.py
```diff
@@ -299,6 +299,173 @@
...
+    # from 16-231. Since the RGB value of the "standard" 256 color
+    # palette can be easily calculated, we can emulate them in RGB
+    # mode rather easily
+    num -= 16
+    return rgb_alloc(_rgb_from_6cube((num / 36) % 6, (num / 6) % 6, num % 6))
```
ulfalizer: *`_rgb_alloc()`

Also need to use `//` instead of `/`, because it ends up as floating-point division otherwise. That causes a crash in `init_color()` for e.g. `MENUCONFIG_STYLE='separator=bg:208'`, because one of the arguments ends up `> 1000`.

menuconfig.py
```diff
@@ -299,6 +299,173 @@
...
+
+def _color_get_rgb(rgb):
+    # Lambda for calculating the Euclidean distance between two RGB colors
+    dist = lambda r1, r2: sum((x - y)**2 for x, y in zip(r1, r2))
+
+    # Best case -- terminal supports the changing of colors
+    if curses.COLORS >= 256 and curses.can_change_color():
+        return rgb_alloc(rgb)
```
ulfalizer: *`_rgb_alloc()`

menuconfig.py
```diff
@@ -299,6 +299,173 @@
...
+
+    # Map each RGB component from the range 0-255 to the range 0-1000, which
+    # is what curses uses
+    curses.init_color(cn, *(int(round(x * 1000 / 255)) for x in rgb))
+    rgb2index[rgb] = cn
+
+    return cn
+
+def _color_get(num):
+    """
```
ulfalizer: For internal functions, the convention is to use "doc comments" instead of docstrings. See e.g. `_is_num()`.

Ditto for other functions.

menuconfig.py
```diff
@@ -299,6 +299,173 @@
...
+    # mode rather easily
+    num -= 16
+    return rgb_alloc(_rgb_from_6cube((num / 36) % 6, (num / 6) % 6, num % 6))
+
+def _color_get_rgb(rgb):
+    # Lambda for calculating the Euclidean distance between two RGB colors
+    dist = lambda r1, r2: sum((x - y)**2 for x, y in zip(r1, r2))
+
+    # Best case -- terminal supports the changing of colors
+    if curses.COLORS >= 256 and curses.can_change_color():
```
ulfalizer: Maybe the `if curses.COLORS >= 256` could be factored out:

```python
if curses.COLORS >= 256:
    ...
    if curses.can_change_color():
        ...
    else:
        ...
```

pinkfluid: Maybe we can change it to curses.COLORS >= 16, as the algorithm assumes we should not change the first 16 colors.

pinkfluid: Good catch at the division error. Fixed in the new version.

pinkfluid:
I believe I addressed the last batch of change requests.

pinkfluid:
I allowed myself to add another feature (it's 3 lines of code, so not sure if its worth a whole pull request).

Basically, you copy styles from other elements, like this:

MENUCONFIG_STYLE="separator=fg:white,bg:red path=separator"

ulfalizer:
> I allowed myself to add another feature (it's 3 lines of code, so not sure if its worth a whole pull request).
> 
> Basically, you copy styles from other elements, like this:
> 
> MENUCONFIG_STYLE="separator=fg:white,bg:red path=separator"

No objections from me. Seems like a nice feature.

Will take a look after I'm done with work for today. Some documentation for the new color format and the `path=separator` feature in the module docstring would be nice as well.

pinkfluid:
Pushed a new squashed commit with docstring updates for the new color format and the path=separator feature.

menuconfig.py
ulfalizer: > Maybe we can change it to curses.COLORS >= 16, as the algorithm assumes we should not change the first 16 colors.

Hmm... maybe a higher limit would be nice though, to avoid running out of palette entries. Wonder if you ever run across any terminals with 16 < COLORS < 256.

ulfalizer:
Seems pretty robust now from playing around with it a bit. Merged.

Thanks! Learned some things about terminal colors.

PR #72, opened 2019-07-13
Fix the 'source "$ENV/xx/yy"' not found issue when defined the 'ENV' …

armink:
…by 'config ENV'.

testcase like this:

```
config RTT_DIR
    string
    option env="RTT_ROOT"
    default "../../.."

source "$RTT_DIR/Kconfig"
```

### DIFF
--- a/kconfiglib.py
+++ b/kconfiglib.py
@@ -2822,6 +2822,12 @@ class Kconfig(object):
 
                 self._parse_properties(node)
 
+                if node.item.env_var:
+                    if node.item.env_var in os.environ:
+                        os.environ[node.item.name] = os.environ[node.item.env_var]
+                    else:
+                        os.environ[node.item.name] = ((node.defaults[0])[0]).name
+
                 if node.is_menuconfig and not node.prompt:
                     self._warn("the menuconfig symbol {} has no prompt"
                                .format(_name_and_loc(sym)))


ulfalizer:
Hello,

`option env` symbols are only supported in a limited way for backwards compatibility. They're accepted, but ignored, which is where that same-name limitation comes from.

If you need a level of indirection with the naming, you can do this instead, using the preprocessor:

```
# RTT_ROOT does not exist as a preprocessor variable, so the value is fetched
# from the environment
RTT_DIR := $(RTT_ROOT)

...

source "$(RTT_DIR)/Kconfig"
```

It's a bit rude for the library to directly modify `os.environ`, so I'd prefer not to do that.

See this note in the README:

    Environment variables are expanded directly in e.g. source and mainmenu
    statements, meaning option env symbols are redundant.

    This is the standard behavior with the new Kconfig preprocessor, which
    Kconfiglib implements.

    'option env' symbols are supported for backwards compatibility, with the caveat
    that they must have the same name as the environment variables they reference.
    A warning is printed if the names differ.

armink:
Can I add a Kconfig standard version parameter for compatible with old project? It's maybe set by `KCONFIG_OLD_STANDARD` or `KCONFIG_STANDARD VERSION` env.

ulfalizer:
I prefer to not add behavior tweak flags. Usually means they have to be supported for a long time, and they complicate the code.

Going through the Kconfig files and converting them to the preprocessor might be the best approach. If there's unknown third-party stuff, then maybe a fork could be used for a while, that adds a deprecation warning for a release or two before you switch over.

ulfalizer:
I should probably start generating deprecation warning for `option env` in Kconfiglib itself too at some point. It might be hard to discover that there's a simpler way to do things now.

armink:
If the old version is compatible with minor changes, it may be allowed the library to get more users. :)

ulfalizer:
In most projects, getting rid of `option env` just takes a few minutes.

The C tools have dropped `option env` by the way, so it's more compatible than those. The reason I decided to keep supporting `option env` in a limited form is that it's nice to be able to run Kconfiglib on older Linux kernels. There's also a minor `help` parsing hack to keep supporting older kernels.

The kernel always named the Kconfig symbol the same as the environment variables, so the require-same-name hack is enough to keep it working for those.

ulfalizer:
Added a commit that clarifies why I decided to keep a limited version of `option env`: https://github.com/ulfalizer/Kconfiglib/commit/a2afc4d34d61bc8bf0e005112256fdc064cf1beb

Might see why I think it's a bit icky to extend support for it. There doesn't seem to be a clean way to do it either. At some point I want to drop it, or at least start generating deprecation warnings for it. The code overhead isn't that great though.

ulfalizer:
Closing for now.

Feel free to open another issue/PR if something comes up.

Issue #71, opened 2019-06-20
how to merge multiple Kconfig(absolute path) by  `source "${var}"`


Neutree:
I want to merge three `Kconfig` file to one,

* root Kconfig(abs path: `/home/aaa/Kconfig`):
```
mainmenu "C/CPP CMake project framework Kconfig configuration"
menu "Components configuration"
	source "$(SUB_KCONFIGS)"
endmenu
```
* Kconfig 1(abs path: `/home/aaa/abcd/Kconfig`):
```
...
```

* Kconfig 2(abs path: `/root/aaa/abcd/Kconfig`):
```
...
```

and I want to merge Kconfig 2 and Kconfig 1 to root Kconfig like this way
```
cd /home/aaa/
export SUB_KCONFIGS="/home/aaa/abcd/Kconfig /root/aaa/abcd/Kconfig"
menuconfig
```

but it's not work, maybe because of this code:

https://github.com/ulfalizer/Kconfiglib/blob/f1e6f328bf30a7a88c5a48a863bdd55aed896b1e/kconfiglib.py#L2846-L2856

for temporary use， I changed code to 
```python
filenames = pattern.split(" ")
if not os.path.exists(filenames[0]):
    filenames = sorted(iglob( join(self._srctree_prefix, pattern) ))
```

but can you tell me what's the right way to merge these three kconfig files?

best regards!

ulfalizer:
Hello,

`source` currently isn't designed to include more than one file, except via globbing.

To implement it, you'd have to do something like this, so that the code works with a list of patterns instead of a single pattern (untested):

```python
elif t0 in _SOURCE_TOKENS:
    patterns = self._expect_str_and_eol().split()

    if t0 in _REL_SOURCE_TOKENS:
        patterns = [join(dirname(self._filename), pattern)
                    for pattern in patterns]

    filenames = []
    for pattern in pattern:
        filenames += glob(join(self,_srctree_prefix, pattern))

    filenames.sort()

    # Same as before...
```

Could add a helper too and do this instead:

```python
elif t0 in _SOURCE_TOKENS:
    for pattern in self._expect_str_and_eol().split():
        _source_pattern(t0, pattern)
```

`_source_pattern()` would be the same as the old logic.

How are you planning to use it by the way?

One thing that makes it feel a bit iffy to me is that the Kconfig `source` requires its argument to be quoted. Would be nice if it had been designed to work like `include` in `make` from the beginning, where you can do this:

```
[...]

PR #75, opened 2019-09-18
Add "lsource" for sourcing a list of files

dobairoland:
Here is the PR I mentioned here: https://github.com/ulfalizer/Kconfiglib/issues/71#issuecomment-532184268

I hope it is worth of adding to your great library. I'm ready to make any necessary adjustments.

### DIFF
--- a/kconfiglib.py
+++ b/kconfiglib.py
@@ -420,6 +420,28 @@ files matching "bar*" exist:
 
 'source' and 'osource' are analogous to 'include' and '-include' in Make.
 
+Listing 'source'
+----------------
+
+(This is a provisional feature and might change. The major version of the
+library will be increased if any changes are made.)
+
+'lsource' accepts a list of patterns for sourcing. For example,
+
+  lsource "foo bar"
+
+is equivalent to
+
+  source "foo"
+  source "bar"
+
+When combined with the Kconfig preprocessor, this can be used to include a list
+of files from an environment variable, like
+
+  lsource "$(ENV_VAR)"
+
+A preprocessor function could be used as well.
+
 
 Generalized def_* keywords
 --------------------------
@@ -2857,15 +2879,18 @@ class Kconfig(object):
                     # Relative source
                     pattern = join(dirname(self.filename), pattern)
 
-                # - glob() doesn't support globbing relative to a directory, so
-                #   we need to prepend $srctree to 'pattern'. Use join()
-                #   instead of '+' so that an absolute path in 'pattern' is
-                #   preserved.
-                #
-                # - Sort the glob results to ensure a consistent ordering of
-                #   Kconfig symbols, which indirectly ensures a consistent
-                #   ordering in e.g. .config files
-                filenames = sorted(iglob(join(self._srctree_prefix, pattern)))
+                if t0 is _T_LSOURCE:
+                    filenames = [join(self._srctree_prefix, i) for i in pattern.split()]
+                else:
+                    # - glob() doesn't support globbing relative to a directory, so
+                    #   we need to prepend $srctree to 'pattern'. Use join()
+                    #   instead of '+' so that an absolute path in 'pattern' is
+                    #   preserved.
+                    #
+                    # - Sort the glob results to ensure a consistent ordering of
+                    #   Kconfig symbols, which indirectly ensures a consistent
+                    #   ordering in e.g. .config files
+                    filenames = sorted(iglob(join(self._srctree_prefix, pattern)))
 
                 if not filenames and t0 in _OBL_SOURCE_TOKENS:
                     raise KconfigError(
@@ -6748,6 +6773,7 @@ except AttributeError:
     _T_INT,
     _T_LESS,
     _T_LESS_EQUAL,
+    _T_LSOURCE,
     _T_MAINMENU,
     _T_MENU,
     _T_MENUCONFIG,
@@ -6769,7 +6795,7 @@ except AttributeError:
     _T_TRISTATE,
     _T_UNEQUAL,
     _T_VISIBLE,
-) = range(1, 51)
+) = range(1, 52)
 
 # Keyword to token map, with the get() method assigned directly as a small
 # optimization
@@ -6800,6 +6826,7 @@ _get_keyword = {
     "if":             _T_IF,
     "imply":          _T_IMPLY,
     "int":            _T_INT,
+    "lsource":        _T_LSOURCE,
     "mainmenu":       _T_MAINMENU,
     "menu":           _T_MENU,
     "menuconfig":     _T_MENUCONFIG,
@@ -6896,6 +6923,7 @@ _STRING_LEX = frozenset({
     _T_COMMENT,
     _T_HEX,
     _T_INT,
+    _T_LSOURCE,
     _T_MAINMENU,
     _T_MENU,
     _T_ORSOURCE,
@@ -6923,6 +6951,7 @@ _SOURCE_TOKENS = frozenset({
     _T_RSOURCE,
     _T_OSOURCE,
     _T_ORSOURCE,
+    _T_LSOURCE,
 })
 
 _REL_SOURCE_TOKENS = frozenset({
@@ -6934,6 +6963,7 @@ _REL_SOURCE_TOKENS = frozenset({
 _OBL_SOURCE_TOKENS = frozenset({
     _T_SOURCE,
     _T_RSOURCE,
+    _T_LSOURCE,
 })
 
 _BOOL_TRISTATE = frozenset({


kconfiglib.py
```diff
@@ -420,6 +420,18 @@
...
+
+'lsource' accepts a list of patterns for sourcing. For example,
+
+  lsource "foo bar"
+
+is equivalent to
+
+  source "foo"
+  source "bar"
+
```
ulfalizer: Could add something like

```
When combined with the Kconfig preprocessor, this can be used to include a list
of files from an environment variable, like

  lsource "$(ENV_VAR)"

A preprocessor function could be used as well.
```

kconfiglib.py
```diff
@@ -6769,7 +6784,8 @@ def _shell_fn(kconf, _, command):
     _T_TRISTATE,
     _T_UNEQUAL,
     _T_VISIBLE,
-) = range(1, 51)
+    _T_LSOURCE,
```
ulfalizer: List is sorted alphabetically.

kconfiglib.py
```diff
@@ -6817,6 +6833,7 @@ def _shell_fn(kconf, _, command):
     "string":         _T_STRING,
     "tristate":       _T_TRISTATE,
     "visible":        _T_VISIBLE,
+    "lsource":        _T_LSOURCE,
```
ulfalizer: Sorted alphabetically.

kconfiglib.py
```diff
@@ -6905,6 +6922,7 @@ def _shell_fn(kconf, _, command):
     _T_SOURCE,
     _T_STRING,
     _T_TRISTATE,
+    _T_LSOURCE,
```
ulfalizer: Sorted alphabetically.

kconfiglib.py
```diff
@@ -420,6 +420,18 @@
 
 'source' and 'osource' are analogous to 'include' and '-include' in Make.
 
+Listing 'source'
+----------------
+
```
ulfalizer: Could add a note like

```
(This is a provisional feature and might change. The major version will be increased for any changes.)
```

See the main commit message.

dobairoland:
Thank you @ulfalizer for your suggestions and for the will to work toward this feature. I corrected the commit as per your suggestions.

> It's a bit inconsistent in that there's no relative lsource, but at the same time I don't want to put in too much stuff that would never get used, so that might be fine.

Would you like more the idea that `source`, `rsource` and `osource` would accept a list as well (instead of adding `lsource`)?

ulfalizer:
> Thank you @ulfalizer for your suggestions and for the will to work toward this feature. I corrected the commit as per your suggestions.

Thanks!

> Would you like more the idea that `source`, `rsource` and `osource` would accept a list as well (instead of adding `lsource`)?

One disadvantage is that it breaks paths with spaces in them. Probably very rare to have those, but you never know what people might depend on.

I'm fine with `lsource` for now. Even if I mark it provisional, I won't go changing stuff willy-nilly, because you probably shouldn't make people's lives too painful. In practice, it'll probably stay forever, unchanged. :)

Will look more over the weekend. Sorry for slowness. Should calm down soon.

dobairoland:
> One disadvantage is that it breaks paths with spaces in them. Probably very rare to have those, but you never know what people might depend on.

Yes, poor Windows users. :-) I forgot about those spaces. However, the list items could be separated with something else, e.g. semicolon.

 > Will look more over the weekend.

Thank you for your time!

dobairoland:
I discovered an incompability with Python 2.

https://github.com/ulfalizer/Kconfiglib/blob/master/kconfiglib.py#L2472 token is `unicode` and not `str`, so it fails with `kconfiglib.KconfigError: /home/dragon/esp/esp-idf/Kconfig:64: couldn't parse 'lsource "$COMPONENT_KCONFIGS_PROJBUILD"': expected string`.

ulfalizer:
> I discovered an incompability with Python 2.
> 
> https://github.com/ulfalizer/Kconfiglib/blob/master/kconfiglib.py#L2472 token is `unicode` and not `str`, so it fails with `kconfiglib.KconfigError: /home/dragon/esp/esp-idf/Kconfig:64: couldn't parse 'lsource "$COMPONENT_KCONFIGS_PROJBUILD"': expected string`.

Weird. I wonder how you ended up with `unicode` there. The library deliberately only deals with plain undecoded `str` on Python 2, because `unicode`-ifying everything would get messy and make 2/3 compatibility harder.

Could you check if `os.environ["COMPONENT_KCONFIGS_PROJBUILD"]` is a `unicode` strings? That would explain it in that case, though I'm not sure how it'd happen, unless you're manually setting it.

The recommended syntax nowadays is this btw:

```
lsource "$(COMPONENT_KCONFIGS_PROJBUILD)"
```

That's using the new Kconfig preprocessor. The `()`-less syntax is deprecated and only supported for backwards compatibility.

ulfalizer:
I wonder if `lsource` should be in `_OBL_SOURCE_TOKENS`. Allowing a list of zero files seems more flexible, especially if they might be generated by a preprocessor function.

It should really be `lsource`, `olsource`, `rlsource`, and `orlsource` I guess, but it feels like overkill, and the code footprint for that is pretty large.

@dobairoland 
Are all the components that can appear in https://github.com/espressif/esp-idf/blob/96b96ae244f360c84b71c8c3670e24a1186e9844/Kconfig#L67 already within the esp-idf repo by the way?

In that case, maybe you could do something like this instead:

```
if $(has-component,foo)
source "components/foo/Kconfig"
endif

if $(has-component,bar)
source "component/bar/Kconfig"
endif

...
```

`has-component` would be a Python preprocessor function (see https://github.com/ulfalizer/Kconfiglib/blob/f2ce282eca6537210d83f9f0d2c753c2b280943c/kconfiglib.py#L458). It could be defined like this:

```python
import kconfiglib


def has_component(kconf, _, name):
    components = os.environ.get("ESP_IDF_COMPONENTS")
    if components is None:
        raise kconfiglib.KconfigError(
            "{}:{}: $(has-component) requires the env. var. "
            "ESP_IDF_COMPONENTS to be set to a list of component names"
            .format(kconf.filename, kconf.linenr))

    return "y" if name in components.split() else "n"


functions = {
    "has-component": (has_component, 1, 1)
}
```

Thoughts?

It's a bit more work on the esp-idf side, but on the upside, it makes it easy to `grep` for where stuff comes from. The preprocessor functionality is pretty flexible.

dobairoland:
Hi @ulfalizer. Thank you for taking the time and looking at this again.
 
> Could you check if os.environ["COMPONENT_KCONFIGS_PROJBUILD"] is a unicode string?

Yes, it is unicode. Unicode characters in paths are not supported in Kconfiglib?

> I wonder if lsource should be in _OBL_SOURCE_TOKENS. Allowing a list of zero files seems more flexible, especially if they might be generated by a preprocessor function.

I can remove it from there if you prefer that solution.

> It should really be lsource, olsource, rlsource, and orlsource I guess, but it feels like overkill, and the code footprint for that is pretty large.

Yes, that seems to be too much.

> Are all the components that can appear in https://github.com/espressif/esp-idf/blob/96b96ae244f360c84b71c8c3670e24a1186e9844/Kconfig#L67 already within the esp-idf repo by the way?

Unfortunately, no they are not. There are other frameworks which depend on ESP-IDF which has their components. Our customers have their own private components. I cannot even compile a complete list. 

I've tried to add `$(source-list,"$(COMPONENT_KCONFIGS)")` istead of the proposed `lsource` and created `kconfigfunctions.py`:

```python
def source_list(kconf, name, path_list):
    return '\n'.join(['source "{}"'.format(path) for path in path_list[1:-1].split()])

functions = {
    "source-list": (source_list, 1, 1),
}
```
The function returns:
```
source "/home/xy/esp/esp-idf/components/app_update/Kconfig.projbuild"
source "/home/xy/esp/esp-idf/components/bootloader/Kconfig.projbuild"
...
```

But unfortunately, this isn't working.

ulfalizer:
> > Could you check if os.environ["COMPONENT_KCONFIGS_PROJBUILD"] is a unicode string?
> 
> Yes, it is unicode. Unicode characters in paths are not supported in Kconfiglib?

They are supported in the sense that Kconfiglib just avoids decoding any strings on Python 2, and I haven't seen `unicode` strings show up in `os.environ` before. Even if you do this, it stays as a regular `str`:

```
$ foo=ßö python
>>> import os
>>> os.environ["foo"]
'\xc3\x9f\xc3\xb6'
```

That's why I'm wondering how you ended up with `unicode`. Manually setting `os.environ` entries to `unicode` strings currently isn't supported at least, if that's what's going on.

Curious how it happened, so I can see if it's worth fixing.

> > I wonder if lsource should be in _OBL_SOURCE_TOKENS. Allowing a list of zero files seems more flexible, especially if they might be generated by a preprocessor function.
> 
> I can remove it from there if you prefer that solution.

Might be best if there's only gone be one variant.

> Unfortunately, no they are not. There are other frameworks which depend on ESP-IDF which has their components. Our customers have their own private components. I cannot even compile a complete list.

Yeah, that makes it a bit messier.

Another option might be to generate a Kconfig file with a bunch of `source`s in it, though it isn't super pretty either. Something like `Kconfig.modules`.

Any chance that might work?

> I've tried to add `$(source-list,"$(COMPONENT_KCONFIGS)")` istead of the proposed `lsource` and created `kconfigfunctions.py`:
> 
> ```python
> def source_list(kconf, name, path_list):
>     return '\n'.join(['source "{}"'.format(path) for path in path_list[1:-1].split()])
> 
> functions = {
>     "source-list": (source_list, 1, 1),
> }
> ```
> 
> The function returns:
> 
> ```
> source "/home/xy/esp/esp-idf/components/app_update/Kconfig.projbuild"
> source "/home/xy/esp/esp-idf/components/bootloader/Kconfig.projbuild"
> ...
> ```
> 
> But unfortunately, this isn't working.

Nah, the preprocessor works at the single token level, so that won't work. It was designed like that when it was added to the C tools to prevent too much anarchy, though it'd be semi-handy here.

ulfalizer:
Sorry for stalling, but `lsource` is starting to feel a bit like an oddball feature that wouldn't get used very often. When I put things in, I have to support them forever, and it has to mesh with whatever gets put in in the future.

I'm a bit afraid that someone will come along in the future and say "hey, I need to source a bunch of glob patterns from an environment variable, why isn't this globbing?", or "why isn't there a relative (or required) version of this?" At that point, it's hard to come up with a good reason why not, even though it starts to get into messy feature creep. :)

Maybe generating a `Kconfig.modules` wouldn't be too bad of a hack, if it's just for one spot. Could `osource` it so that it's still possible to run Kconfig without modules if you need to. `Kconfig.modules` would be fully general too, and won't suffer from feature creep like `lsource` might.

FWIW, `Kconfig.modules` is what Zephyr does.

I know you just want to get it in and get on with the work, so sorry about that. I have to make it work with everything that comes after it though, and deal with issues people have with it. :)

dobairoland:
> That's why I'm wondering how you ended up with unicode. Manually setting os.environ entries to unicode strings currently isn't supported at least, if that's what's going on.

It is getting de-serialized from a JSON object: https://github.com/espressif/esp-idf/blob/master/tools/kconfig_new/confgen.py#L224 and `json.load` returns `unicode`.

> Another option might be to generate a Kconfig file with a bunch of sources in it, though it isn't super pretty either. Something like Kconfig.modules.

Thank you, I'll try this option.

> Sorry for stalling, but lsource is starting to feel a bit like an oddball feature that wouldn't get used very often. When I put things in, I have to support them forever, and it has to mesh with whatever gets put in in the future.

I understand your point. Thank you for your suggestions and your work on this library!

ulfalizer:
> It is getting de-serialized from a JSON object: https://github.com/espressif/esp-idf/blob/master/tools/kconfig_new/confgen.py#L224 and `json.load` returns `unicode`.

I'll look into it. Thanks for the report!

Found https://github.com/theskumar/python-dotenv/issues/176, though I can't find anything in the Python 2 `os` module docs that explicitly forbids it.

Do you still depend on Python 2?

dobairoland:
> Do you still depend on Python 2?

Yes, we have to support 2.7 in all of our active releases.

ulfalizer:
Dug around a bit. I think it might be best to encode the strings on the esp-idf side before storing them into `os.environ`:

 - `mypy --py2 foo.py` prints

        foo.py:3: error: Incompatible types in assignment (expression has type "unicode", target has type "str")

   for this script:

   ```python
   import os
   os.environ["foo"] = u"bar"
   ```

   mypy is an official Python thing, so might be a sign it's not intended. It supports allowing more than one type too.

 - Seems `unicode` values in `os.environ` can cause issues with `subprocess` on Windows: https://github.com/theskumar/python-dotenv/pull/101

Maybe doing something similar to what [this PR](https://github.com/theskumar/python-dotenv/pull/177/commits/45f697a7f1c273f304a41d74555223fe85eee528#diff-801425242c4901b4508f6da6283b0100R112) does in ESP-IDF would make sense.

dobairoland:
I solved my issue by creating a temporary file with a list of source item.

Thank you for your help and kind suggestions.

Issue #77, opened 2019-09-27
Kconfiglib cannot be installed from MSYS


dobairoland:
```
$ pip install kconfiglib
Collecting kconfiglib
  Downloading https://files.pythonhosted.org/packages/13/3b/836b3af23701728830771d5c45fc3bd1c67ea646b49f6384a44343aa0b3e/kconfiglib-12.14.1-py2.py3-none-any.whl (142kB)
     |################################| 143kB 2.2MB/s
Collecting windows-curses; sys_platform == "win32" (from kconfiglib)
  ERROR: Could not find a version that satisfies the requirement windows-curses; sys_platform == "win32" (from kconfiglib) (from versions: none)
ERROR: No matching distribution found for windows-curses; sys_platform == "win32" (from kconfiglib)
```

I closed this https://github.com/zephyrproject-rtos/windows-curses/issues/4 accepting that `windows-curses` cannot be installed in MSYS but I think Kconfiglib should not request it there if it cannot be installed.

ulfalizer:
Didn't know it reported as `win32` to be honest. I'll look into it once I get back to Windows box.

dobairoland:
Usually, we test for `MSYSTEM` if it is in `os.environ` but I don't see anything like that here: https://www.python.org/dev/peps/pep-0508/#environment-markers

ulfalizer:
Got MSYS2 installed now.

> Usually, we test for `MSYSTEM` if it is in `os.environ` but I don't see anything like that here: https://www.python.org/dev/peps/pep-0508/#environment-markers

Yeah, I don't see a way to identify MSYS2 with environment markers either. Bit messy that it doesn't distinguish itself from the "official" Python. Would be nice if there was an "install if compatible package available" option in `setuptools`/wheels, but I didn't see anything like that.

Looking at https://github.com/msys2/msys2/wiki/Terminals, MSYS2 doesn't use the regular Windows Console API for its terminal (it uses Mintty), so it won't be compatible with `windows-curses`.

I could drop `windows-curses` from `install_requires` in `setup.py` and require it to be installed manually, but since people might depend on it being automatically installed, it might be a bit mean.

I'll dig some more to see if I can find a way to detect MSYS2 or only install `windows-curses` if possible, but seems messy so far.

[...]

PR #78, opened 2019-10-04
Do not try to install windows-curses in MSYS2

dobairoland:
@ulfalizer Please take a look. I believe that this solves the problem.

Closes https://github.com/ulfalizer/Kconfiglib/issues/77

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,11 @@ setuptools.setup(
     # The terminal menuconfig implementation uses the standard Python 'curses'
     # module. The windows-curses package makes it available on Windows. See
     # https://github.com/zephyrproject-rtos/windows-curses.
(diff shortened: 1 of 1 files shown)


ulfalizer:
Won't work when Kconfiglib is installed as a [Python wheel](https://pythonwheels.com/), unfortunately. Wheels are the usual case nowadays, e.g. when using `pip`.

`setup.py` only gets run when building the wheel, not at installation time.

Thinking I should just drop the `windows-curses` dependency, bump the major version, add a warning somewhere, and hope it doesn't cause too many problems. Would be nice if there was some hacky way to detect MSYS2 with the Wheel format though.

ulfalizer:
Fixed a different way with https://github.com/ulfalizer/Kconfiglib/commit/068bc48cb60bac7b7c241e313b60e7f7707c07f3.

Issue #84, opened 2020-01-02
Menuconfig fails in MacOs with Python 3 when the terminal is resized


dobairoland:
Menuconfig works fine in MacOs with Python 2. However, it fails with Python 3 with the following error when the terminal window is resized:

```
File "/Users/xy/esp/esp-idf/tools/kconfig_new/menuconfig.py", line 3300, in <module>
_main()
File "/Users/xy/esp/esp-idf/tools/kconfig_new/menuconfig.py", line 672, in _main
menuconfig(standard_kconfig(_doc_))
File "/Users/xy/esp/esp-idf/tools/kconfig_new/menuconfig.py", line 741, in menuconfig
print(curses.wrapper(_menuconfig))
File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/curses/_init_.py", line 102, in wrapper
return func(stdscr, *args, **kwds)
File "/Users/xy/esp/esp-idf/tools/kconfig_new/menuconfig.py", line 844, in _menuconfig
c = _getch_compat(_menu_win)}}
File "/Users/xy/esp/esp-idf/tools/kconfig_new/menuconfig.py", line 3171, in _getch_compat
c = win.get_wch()
_curses.error: no input
```

This issue was observed in https://github.com/espressif/esp-idf but probably it can be reproduced without it. We reproduced it on several machines.

The issue cannot be observed in Linux nor Windows. It works there as it is expected.

ulfalizer:
Don't have a Mac handy, unfortunately.

To narrow it down, try running `python3 menuconfig.py <dummy file>` directly outside CMake on this dummy file:

```
config FOO
	bool "foo"
```

Also try running this program:

```
import curses


def main(stdscr):
    while True:
        stdscr.get_wch()


curses.wrapper(main)
```

I wonder if it might be related to [this issue](https://stackoverflow.com/questions/23203735/importing-rlcompleter-causes-terminal-resize-to-fail).

kumekay:
I've tried and both `menuconfig.py` and test program are failing.
Both pyenv installed python 3.8 and xcode's python 3.7 behave the same way:

```
Traceback (most recent call last):
  File "t.py", line 9, in <module>
    curses.wrapper(main)
  File "/Users/ku/.pyenv/versions/3.8.0/lib/python3.8/curses/__init__.py", line 105, in wrapper
    return func(stdscr, *args, **kwds)
[...]

PR #85, opened 2020-01-03
Fix resize of menuconfig in Python 3 on MacOs

dobairoland:
Closes https://github.com/ulfalizer/Kconfiglib/issues/84

### DIFF
--- a/menuconfig.py
+++ b/menuconfig.py
@@ -3128,9 +3128,10 @@ def _getch_compat(win):
     # Uses get_wch() if available (Python 3.3+) and getch() otherwise. Also
     # handles a PDCurses resizing quirk.
 
-    if hasattr(win, "get_wch"):
+    try:
         c = win.get_wch()
-    else:
+        # curses.error is thrown in Python 3 on MacOs
+    except (AttributeError, curses.error):
         c = win.getch()
         if 0 <= c <= 255:
             c = chr(c)


dobairoland:
@ulfalizer @kumekay Please take a look. I think this is a sufficient fix.

kumekay:
This looks good and works for me just fine

menuconfig.py
```diff
@@ -3128,9 +3128,10 @@ def _getch_compat(win):
     # Uses get_wch() if available (Python 3.3+) and getch() otherwise. Also
     # handles a PDCurses resizing quirk.
 
-    if hasattr(win, "get_wch"):
+    try:
         c = win.get_wch()
-    else:
+        # curses.error is thrown in Python 3 on MacOs
```
ulfalizer: Could change the message to say

```
# Work around a crash where get_wch() raises curses.error on macOS when
# resizing the terminal by falling back on getch(). This needs more
# investigation. It doesn't happen on all systems.
```

See https://github.com/ulfalizer/Kconfiglib/issues/84#issuecomment-572450946 too. Would be nice to narrow down the configurations that have this issue.

ulfalizer:
Pushed out a release with a fix (same as this one, just with some more comments): 68bcecd

Would be nice to figure out and report the root issue at some point though.

Thanks.

PR #90, opened 2020-01-24
Change all hashbangs to use python3

### LINKED ISSUE #89: Use Python 3 in scripts
Now that Python 2 is EOL, would it be sensible to change the hashbangs in all the scripts to use `#! /usr/bin/env python3`?

There's an increasing number of systems without Python 2 at all, and they don't *need* to have a `python` binary available.  For example, my Debian chroot for development in doesn't have a `/usr/bin/python` but has Python 3 installed.

rossburton:
Now that Python 2 is pretty much dead, change all script hashbangs to use
/usr/bin/env python3.

Nothing else is changed: the module still supports Python 2, and when installing
with python2 the entry_point bounce scripts have a python2 hashbang.

Closes #89.

### DIFF
--- a/alldefconfig.py
+++ b/alldefconfig.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright (c) 2018-2019, Ulf Magnusson
 # SPDX-License-Identifier: ISC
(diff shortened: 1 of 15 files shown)


ulfalizer:
Heh... was preparing a commit in parallel. Pushed now.

I'll push out a release as well, for people who fetch the source code from PyPI.

ulfalizer:
Pushed out in a new 14.0.0 release.

Thanks for the report!

Issue #93, opened 2020-05-22
menuconfig: Exception when leaving menu of choice with multiple definitions


leandrolanzieri:
## Description
When a choice is given an associated symbol, it can have multiple definitions, so the options could be extended like so:

```
mainmenu "Main"

menu "Sub"
choice MY_CHOICE
    bool "Test choice"

config MY_OPT1
    bool "Option 1"

config MY_OPT2
    bool "Option 2"

endchoice
endmenu

choice MY_CHOICE
    default MY_OPT3

config MY_OPT3
    bool "Option 3"

endchoice
```

While using menuconfig, if one searches for `MY_OPT3` and jumps to the symbol the choice menu will be displayed. If then one tries to leave the menu, an exception is raised, as the parent menu (top menu in this case) does not actually display the current menu.

## Steps to reproduce
1. Using the snippet above, run menuconfig
2. Press `/` ('Jump to symbol') and search for `MY_OPT3`
3. Jump to the symbol (pressing `Enter`)
4. Leave the menu (pressing `Esc`)

## Expected results
The `Sub` menu is displayed.

## Actual results
An exception is raised and the program stops:

```
Traceback (most recent call last):
  File "/home/leandro/.local/bin/menuconfig", line 10, in <module>
    sys.exit(_main())
  File "/home/leandro/.local/lib/python3.7/site-packages/menuconfig.py", line 631, in _main
    menuconfig(standard_kconfig())
  File "/home/leandro/.local/lib/python3.7/site-packages/menuconfig.py", line 700, in menuconfig
    print(curses.wrapper(_menuconfig))
  File "/usr/lib/python3.7/curses/__init__.py", line 94, in wrapper
    return func(stdscr, *args, **kwds)
  File "/home/leandro/.local/lib/python3.7/site-packages/menuconfig.py", line 859, in _menuconfig
    _leave_menu()
  File "/home/leandro/.local/lib/python3.7/site-packages/menuconfig.py", line 1174, in _leave_menu
    _sel_node_i = _shown.index(_cur_menu)
ValueError: <menu node for choice MY_CHOICE, deps y, has child, Kconfig:21> is not in list
```

PR #94, opened 2020-05-22
menuconfig: Avoid crashing when leaving a menu not shown in parent

leandrolanzieri:
This adds a catch to an exception that might occur when leaving a menu which is not shown by the parent menu. For an example of this refer to issue #93.

For the snippet given in #93 this ends up showing the top menu. I'm not sure if this is the expected behaviour (after all, the second definition of the choice is defined in the top menu), or 'Sub' menu should be shown. At least this seems to avoid the crash.

### DIFF
--- a/menuconfig.py
+++ b/menuconfig.py
@@ -1202,7 +1202,12 @@ def _leave_menu():
     # Jump to parent menu
     parent = _parent_menu(_cur_menu)
     _shown = _shown_nodes(parent)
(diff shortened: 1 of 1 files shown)


jackrosenthal:

:arrow_right: **Action required:** @leandrolanzieri  Are you able to re-open your pull request on the new repository?

leandrolanzieri:
@jackrosenthal see https://github.com/zephyrproject-rtos/Kconfiglib/pull/4

PR #100, opened 2020-12-08
Support printing negatives in Kconfig.write_autoconf()

jackrosenthal:
Some forks of the C implementation of Kconfig (notably, used in
coreboot and depthcharge) offer an option to include false values
defined to zero in generated header files.

This adds support for an equivalent feature to Kconfiglib, so it can
be made use of in those codebases.

### DIFF
--- a/kconfiglib.py
+++ b/kconfiglib.py
@@ -1414,7 +1414,7 @@ class Kconfig(object):
         """
         load_allconfig(self, filename)
 
-    def write_autoconf(self, filename=None, header=None):
+    def write_autoconf(self, filename=None, header=None, negatives=None):
         r"""
         Writes out symbol values as a C header file, matching the format used
         by include/generated/autoconf.h in the kernel.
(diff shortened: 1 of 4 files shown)


kconfiglib.py
```diff
@@ -1445,19 +1445,32 @@ def write_autoconf(self, filename=None, header=None):
...
+
         Returns a string with a message saying that the header got saved, or
         that there were no changes to it. This is meant to reduce boilerplate
         in tools, which can do e.g. print(kconf.write_autoconf()).
         """
         if filename is None:
             filename = os.getenv("KCONFIG_AUTOHEADER",
                                  "include/generated/autoconf.h")
+        if negatives is None:
+            negatives = "KCONFIG_NEGATIVES" in os.environ
```
ismell: Should we check if `KCONFIG_NEGATIVES=1`?

jackrosenthal: Sure, technically the C impl just checks if present:

```
if (getenv("KCONFIG_NEGATIVES")) { ...
```

But since this wouldn't break coreboot/depthcharge behavior to check for `1`, seems like a more intuitive choice.

jackrosenthal:
@ulfalizer can you review?

PR #101, opened 2021-01-27
kconfig: kconfiglib.py: Backup files only

domo141:
The _save_old() to return early if <filename> is not file (or symlink
to a file).

This is simplest alternative to avoid attempt to rename /dev/null
(which could succeed).

This also keeps fifos (perhaps nonexistent potential usage but
this is nicer).

If <filename> were directory or socket, after shutil.copyfile(),
writing to the file (by caller, _write_config()), would fail.

Co-authored-by: Marti Bolivar <marti.bolivar@nordicsemi.no>
Signed-off-by: Tomi Ollila <tomi.ollila@iki.fi>

### DIFF
--- a/kconfiglib.py
+++ b/kconfiglib.py
@@ -1545,8 +1545,7 @@ class Kconfig(object):
           written.
 
           Errors are silently ignored if <filename>.old cannot be written (e.g.
-          due to being a directory, or <filename> being something like
-          /dev/null).
+          due to permissions errors).
 
         verbose (default: None):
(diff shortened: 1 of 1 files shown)


domo141:
See PR with same content in

https://github.com/zephyrproject-rtos/zephyr/pull/31438

and related issue in

https://github.com/zephyrproject-rtos/zephyr/issues/31362

for reasoning

jackrosenthal:

:arrow_right: **Action required:** @domo141   Are you able to re-open your pull request on the new repository?

PR #107, opened 2021-05-30
menuconfig: fix crash caused by unsupported locales

Iksas:
When run on a system with an unsupported locale setting (e.g. `LC_TIME=en_DE.UTF-8`), menuconfig.py crashes with the following message:

    locale.Error: unsupported locale setting

This fixes it by using the default locale if an unsupported locale is encountered.

### DIFF
--- a/menuconfig.py
+++ b/menuconfig.py
@@ -703,8 +703,12 @@ def menuconfig(kconf):
     # errors ourselves.
     kconf.warn = False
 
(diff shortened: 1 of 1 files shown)


jackrosenthal:

:arrow_right: **Action required:** @Iksas  Are you able to re-open your pull request on the new repository?

Iksas:
@jackrosenthal Will do, thanks for the info!

